Issue #7, opened 2017-10-17
Limit memory use for huge dictionary


gendx:
The current implementation of `LZBuffer` allocates the full dictionary size in the constructor, even if the decompressed file is much smaller, which can slow down decompression for small files. Use a dynamic strategy to grow the dictionary on-demand instead, or check the decompressed size in the header (if provided).

gendx:
Added some tests in 43d806c4474491d30cd36709afb2961362978046, it seems that `vec![0; size]` already grows on-demand when data is accessed (at least on Linux). But this may not be the case on other platforms, and the allocation could fail if there is not enough RAM, so a fix is still welcome.

gendx:
Fixed by #22.

PR #8, opened 2017-11-15
Add roundtrip fuzzer

killercup:
Saw [your tweet](https://twitter.com/GEndignoux/status/930760532973547520) and had to play with this a bit! Another interesting fuzzer might be to run another lzma impl and compare the outputs from compress/decompress.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,22 @@
+
+[package]
+name = "lzma-fuzz"
+version = "0.0.1"
+authors = ["Automatically generated"]
+publish = false
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies.lzma]
+path = ".."
+[dependencies.libfuzzer-sys]
+git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "roundtrip"
+path = "fuzz_targets/roundtrip.rs"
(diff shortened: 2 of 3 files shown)


gendx:
Looks awesome, thanks! For now the encoder is very simple so the round-trip may not hit all code paths. It could be nice to fuzz the decoder itself, e.g. starting with a file that decodes as "hello world" https://github.com/gendx/lzma-rs/blob/master/tests/lzma.rs#L100

To compare with other implementations, I guess the popular binding to liblzma from @alexcrichton https://crates.io/crates/lzma-sys is a good target, although I may need to implement .xz first.

killercup:
The roundtrip fuzzer catches exactly 1000 code paths. Adding a decode-only fuzzer is quite simple (just removing lines, basically, and adding it to the Cargo.toml):

```rust
#![no_main]
#[macro_use] extern crate libfuzzer_sys;
extern crate lzma;

fuzz_target!(|data: &[u8]| {
    let mut decomp: Vec<u8> = Vec::with_capacity(data.len());
    let _ = lzma::decompress(&mut std::io::BufReader::new(data), &mut decomp);
});
```

Running this with `cargo fuzz run --release decode -- -rss_limit_mb=8072000000 -max_len=99999` (I got OOM errors without the first flag, and wanted to generate big inputs) quickly gets up to a coverage of around 950 code paths. The fuzzer-magic is able to determine which inputs trigger new code paths and generate more of these interesting inputs. After running this for a while, the `fuzz/corpus/decode/` directory should contain a few valid LZMA files.

But let's help the fuzzer a bit: `cp tests/files/*.lzma fuzz/corpus/decode/`. After this, it starts with 1007 code paths. Running it  for a few minutes makes it go up to 1011 (with diminishing returns).

gendx:
That's great thanks! The OOM errors may be due to #7, I guess it depends on the OS.

For the decoding corpus, maybe smaller files would be more useful. The current test file is 54KB of compressed LMZA, but I have smaller tests inlined directly in the test source code.

I started working on LZMA2 decoding, but I will definitely add more fuzz targets after that!

killercup:
> The OOM errors may be due to #7, I guess it depends on the OS.

It only occurs int he fuzzer context and works when run as a regular test. Probably something with how libfuzzer does things.

> For the decoding corpus, maybe smaller files would be more useful.

Probably, yeah. It'll increase the execs/sec and will yield results faster.

> I started working on LZMA2 decoding

Awesome! 🎉

PR #17, opened 2019-12-09
Add support for specifying the unpacked size outside of header

dragly:
Some LZMA streams, such as those used in OpenCTM, are encoded without
the unpacked size specified in the header. This is possible to read in
some of the C implementations of LZMA by specifying a header size and
providing the unpacked size as an option to the decoder.

This change adds the same possibility to lzma_rs in a typesafe manner,
where the unpacked size and whether it should be written to the header
is specified explicitly.

### Pull Request Overview

This pull request adds `Options` objects to two new public modules named `compress` and `decompress` that can be used to specify whether the unpacked size should be written to and read from the LZMA header.

### Testing Strategy

This exotic use of the LZMA header is only indicated in the [OpenCTM specification itself](http://openctm.sourceforge.net/media/FormatSpecification.pdf), where it specifically states that the offset of the stream [is only 9 bytes](https://github.com/Danny02/OpenCTM/blob/243a343bd23bbeef8731f06ed91e3996604e1af4/doc/FormatSpecification.tex#L91), while it should have been 17 bytes with the unpacked size in place. This is based on 4 bytes from OpenCTM itself, 1 byte for the props, 4 bytes for the dict size and the missing 8 bytes for the unpacked size.

It can also be seen from the OpenCTM source code that the LZMA header written is only 5 bytes:

https://github.com/Danny02/OpenCTM/blob/243a343bd23bbeef8731f06ed91e3996604e1af4/lib/stream.c#L311

### TODO or Help Wanted

This pull request still needs a bit of bikeshedding on the names of the modules and enums used in the options :)

### DIFF
--- a/src/decode/options.rs
+++ b/src/decode/options.rs
@@ -1,4 +1,4 @@
-#[derive(Default)]
+#[derive(Clone, Debug, Default)]
 pub struct Options {
     /// Defines whether the unpacked size should be read from the header or provided.
     /// The default is
@@ -7,6 +7,7 @@ pub struct Options {
 }
 
 /// Alternatives for defining the unpacked size of the decoded data
+#[derive(Clone, Debug)]
 pub enum UnpackedSize {
     /// Assume that the 8 bytes used to specify the unpacked size are present in the header.
     /// If the bytes are `0xFFFF_FFFF_FFFF_FFFF`, assume that there is an end-of-payload marker in
--- a/src/decode/options.rs
+++ b/src/decode/options.rs
@@ -1,7 +1,8 @@
 #[derive(Default)]
 pub struct Options {
     /// Defines whether the unpacked size should be read from the header or provided.
-    /// The default is [`UnpackedSize::ReadHeaderAndUseHeader`](enum.UnpackedSize.html#variant.ReadHeaderAndUseHeader)
+    /// The default is
+    /// [`UnpackedSize::ReadFromHeader`](enum.UnpackedSize.html#variant.ReadFromHeader)
     pub unpacked_size: UnpackedSize,
 }
 
@@ -11,22 +12,23 @@ pub enum UnpackedSize {
     /// If the bytes are `0xFFFF_FFFF_FFFF_FFFF`, assume that there is an end-of-payload marker in
     /// the file.
     /// If not, read the 8 bytes as a little-endian encoded u64.
-    ReadHeaderAndUseHeader,
+    ReadFromHeader,
     /// Assume that there are 8 bytes representing the unpacked size present in the header.
     /// Read it, but ignore it and use the provided value instead.
     /// If the provided value is `None`, assume that there is an end-of-payload marker in the file.
     /// Note that this is a non-standard way of reading LZMA data,
-    /// but is used by certain libraries such as OpenCTM.
+    /// but is used by certain libraries such as
+    /// [OpenCTM](http://openctm.sourceforge.net/).
     ReadHeaderButUseProvided(Option<u64>),
     /// Assume that the 8 bytes typically used to represent the unpacked size are *not* present in
     /// the header. Use the provided value.
     /// If the provided value is `None`, assume that there is an end-of-payload marker in the file.
-    SkipHeaderAndUseProvided(Option<u64>),
+    UseProvided(Option<u64>),
 }
 
 impl Default for UnpackedSize {
     fn default() -> UnpackedSize {
-        UnpackedSize::ReadHeaderAndUseHeader
+        UnpackedSize::ReadFromHeader
     }
 }
 
--- a/src/decode/options.rs
+++ b/src/decode/options.rs
@@ -1,4 +1,4 @@
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default)]
 pub struct Options {
     /// Defines whether the unpacked size should be read from the header or provided.
     /// The default is
@@ -7,7 +7,7 @@ pub struct Options {
 }
 
 /// Alternatives for defining the unpacked size of the decoded data
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
 pub enum UnpackedSize {
     /// Assume that the 8 bytes used to specify the unpacked size are present in the header.
     /// If the bytes are `0xFFFF_FFFF_FFFF_FFFF`, assume that there is an end-of-payload marker in
--- a/src/decode/options.rs
+++ b/src/decode/options.rs
@@ -1,0 +1,32 @@
+#[derive(Default)]
+pub struct Options {
+    /// Defines whether the unpacked size should be read from the header or provided.
+    /// The default is [`UnpackedSize::ReadHeaderAndUseHeader`](enum.UnpackedSize.html#variant.ReadHeaderAndUseHeader)
+    pub unpacked_size: UnpackedSize,
+}
+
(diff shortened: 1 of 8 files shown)


src/decode/options.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub unpacked_size: UnpackedSize,
+}
+
+/// Alternatives for defining the unpacked size of the decoded data
+pub enum UnpackedSize {
+    /// Assume that the 8 bytes used to specify the unpacked size are present in the header.
+    /// If the bytes are `0xFFFF_FFFF_FFFF_FFFF`, assume that there is an end-of-payload marker in
+    /// the file.
+    /// If not, read the 8 bytes as a little-endian encoded u64.
+    ReadHeaderAndUseHeader,
```
gendx: `ReadFromHeader` would be simpler.

src/decode/options.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// Assume that there are 8 bytes representing the unpacked size present in the header.
+    /// Read it, but ignore it and use the provided value instead.
+    /// If the provided value is `None`, assume that there is an end-of-payload marker in the file.
+    /// Note that this is a non-standard way of reading LZMA data,
+    /// but is used by certain libraries such as OpenCTM.
```
gendx: Can you add a link to OpenCTM here?

src/decode/options.rs
```diff
@@ -0,0 +1,32 @@
...
+    ReadHeaderButUseProvided(Option<u64>),
+    /// Assume that the 8 bytes typically used to represent the unpacked size are *not* present in
+    /// the header. Use the provided value.
+    /// If the provided value is `None`, assume that there is an end-of-payload marker in the file.
+    SkipHeaderAndUseProvided(Option<u64>),
```
gendx: `UseProvided` would be simpler.

src/encode/options.rs
```diff
@@ -0,0 +1,28 @@
+/// Options for the `lzma_compress` function
+#[derive(Default)]
+pub struct Options {
+    /// Defines whether the unpacked size should be written to the header and whether the value is
+    /// known.
```
gendx: IMO "and whether the value is known" is not necessary here.

src/encode/options.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// The default is
+    /// [`UnpackedSize::WriteToHeader(None)`](enum.encode.UnpackedSize.html#variant.WriteValueToHeader)
+    pub unpacked_size: UnpackedSize,
+}
+
+/// Alternatives for handling unpacked size
+pub enum UnpackedSize {
+    /// If the value is `Some(u64)`, write the provided u64 value to the header.
+    /// There is currently no check in place that verifies that this is the actual number of bytes provided by the input stream.
```
gendx: Nit: this line overflows by too many characters to the right.

src/encode/options.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// If the value is `None`, write the special `0xFFFF_FFFF_FFFF_FFFF` code to the header,
+    /// indicating that the unpacked size is unknown.
+    WriteToHeader(Option<u64>),
+    /// Do not write anything to the header. The unpacked size needs to be stored elsewhere and
+    /// provided when reading the file. Note that this is a non-standard way of writing LZMA data,
+    /// but is used by certain libraries such as OpenCTM.
```
gendx: Same as above, can you link to OpenCTM?

src/lib.rs
```diff
@@ -10,11 +10,29 @@ pub mod error;
...
+/// Decompress LZMA data with default [`Options`](decompress/struct.Options.html).
 pub fn lzma_decompress<R: io::BufRead, W: io::Write>(
     input: &mut R,
     output: &mut W,
 ) -> error::Result<()> {
-    let params = decode::lzma::LZMAParams::read_header(input)?;
+    lzma_decompress_with_options(input, output, &decompress::Options::default())
+}
+
+/// Decompress LZMA data with the provided options
```
gendx: Nit: missing period at end of sentence.

src/lib.rs
```diff
@@ -10,11 +10,29 @@ pub mod error;
...
+pub fn lzma_decompress_with_options<R: io::BufRead, W: io::Write>(
+    input: &mut R,
+    output: &mut W,
+    options: &decompress::Options,
+) -> error::Result<()> {
+    let params = decode::lzma::LZMAParams::read_header(input, &options)?;
```
gendx: `options` is already a reference. No need to borrow it.

tests/lzma.rs
```diff
@@ -113,3 +133,61 @@ fn decompress_huge_dict() {
...
+}
+
+#[test]
+fn unpacked_size_provided_outside() {
+    let data = b"Some data";
+    let encode_options = lzma_rs::compress::Options {
+        unpacked_size: lzma_rs::compress::UnpackedSize::SkipWritingToHeader,
+    };
+    let decode_options = lzma_rs::decompress::Options {
+        unpacked_size: lzma_rs::decompress::UnpackedSize::SkipHeaderAndUseProvided(Some(
```
gendx: It seems that `SkipHeaderAndUseProvided(None)` isn't tested.

tests/lzma.rs
```diff
@@ -113,3 +133,61 @@ fn decompress_huge_dict() {
...
+#[test]
+fn unpacked_size_write_some_to_header_but_use_provided_on_read() {
+    let data = b"Some data";
+    let encode_options = lzma_rs::compress::Options {
+        unpacked_size: lzma_rs::compress::UnpackedSize::WriteToHeader(
+            Some(data.len() as u64),
+        ),
+    };
+    let decode_options = lzma_rs::decompress::Options {
+        unpacked_size: lzma_rs::decompress::UnpackedSize::ReadHeaderButUseProvided(Some(
```
gendx: It seems that `ReadHeaderButUseProvided(None)` isn't tested.

dragly:
Thanks for the comments!

After writing the tests, I realized that the end-of-stream marker is always written. However, this is documented in XZ as only being supposed to be written if it is not provided in the header.

We should probably not write this if the value is provided and written in the header. Is removing the bits encoded in Encoder::finish the way to achieve this?

https://github.com/gendx/lzma-rs/blob/4d46835872db451886276e7e0681ce927c6348b6/src/encode/dumbencoder.rs#L71-L97

Not sure exactly what should be done if it is known, but not provided, like in the case of OpenCTM. I will see if I can figure out what happens in other libs in that case.

gendx:
> Thanks for the comments!
> 
> After writing the tests, I realized that the end-of-stream marker is always written. However, this is documented in XZ as only being supposed to be written if it is not provided in the header.
> 
> We should probably not write this if the value is provided and written in the header. Is removing the bits encoded in Encoder::finish the way to achieve this?
> 
> https://github.com/gendx/lzma-rs/blob/4d46835872db451886276e7e0681ce927c6348b6/src/encode/dumbencoder.rs#L71-L97
> 
> Not sure exactly what should be done if it is known, but not provided, like in the case of OpenCTM. I will see if I can figure out what happens in other libs in that case.

As the name goes, the `dumbencoder` is currently quite dumb. But I guess removing that part of the `Encoder::finish` would work. You can also test the result against `unlzma` in the command line to see if that works.

dragly:
Seems like it works fine with unlzma, so I removed writing the end marker if the unpacked size is written to the header.

I also realized that a couple of the tests did not really make sense after this and removed those.

PR #22, opened 2019-12-19
Do not allocate buffer size before necessary

dragly:
This changes the way the LZCircularBuffer is allocated. Instead of allocating the full dict size immediately, the buffer is grown on demand. This helps when parsing smaller files with large dict sizes, especially when compiling to WebAssembly and running in browser where `alloc_zeroed` is slow.

### DIFF
--- a/src/decode/lzbuffer.rs
+++ b/src/decode/lzbuffer.rs
@@ -138,7 +138,7 @@ where
     pub fn from_stream(stream: &'a mut W, dict_size: usize) -> Self {
         info!("Dict size in LZ buffer: {}", dict_size);
         Self {
-            stream: stream,
+            stream,
             buf: Vec::new(),
             dict_size,
             cursor: 0,
@@ -147,14 +147,13 @@ where
     }
 
     fn get(&self, index: usize) -> u8 {
-        match self.buf.get(index) {
-            Some(x) => *x,
-            None => 0,
-        }
+        *self.buf.get(index).unwrap_or(&0)
     }
 
     fn get_mut(&mut self, index: usize) -> &mut u8 {
-        self.buf.resize(index + 1, 0);
+        if self.buf.len() < index + 1 {
+            self.buf.resize(index + 1, 0);
+        }
         self.buf.get_mut(index).unwrap()
     }
 }
--- a/src/decode/lzbuffer.rs
+++ b/src/decode/lzbuffer.rs
@@ -138,12 +138,24 @@ where
     pub fn from_stream(stream: &'a mut W, dict_size: usize) -> Self {
         info!("Dict size in LZ buffer: {}", dict_size);
         Self {
-            stream,
-            buf: vec![0; dict_size],
+            stream: stream,
+            buf: Vec::new(),
             dict_size,
             cursor: 0,
             len: 0,
         }
+    }
+
+    fn get(&self, index: usize) -> u8 {
+        match self.buf.get(index) {
+            Some(x) => *x,
+            None => 0,
+        }
+    }
+
+    fn get_mut(&mut self, index: usize) -> &mut u8 {
+        self.buf.resize(index + 1, 0);
+        self.buf.get_mut(index).unwrap()
     }
 }
 
@@ -160,7 +172,7 @@ where
         if self.len == 0 {
             lit
         } else {
-            self.buf[(self.dict_size + self.cursor - 1) % self.dict_size]
+            self.get((self.dict_size + self.cursor - 1) % self.dict_size)
         }
     }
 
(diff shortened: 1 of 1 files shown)


dragly:
We have had performance issues when parsing smaller OpenCTM files using lzma-rs. Profiling showed that the allocation of the LZCircularBuffer was the issue, because memset appears to be pretty slow when running WebAssembly in browsers. In particular, files that take less than a few milliseconds to actually parse will spend and additional 100 ms just to allocate this buffer if the dict size is large. In our case, the dict size is 67M in these files, even though the file itself may be just a few hundred bytes.

dragly:
And I am not sure if this is the cleanest way of achieving dynamic resizing. Let me know if you have any suggestions :)

src/decode/lzbuffer.rs
```diff
@@ -138,13 +138,25 @@ where
     pub fn from_stream(stream: &'a mut W, dict_size: usize) -> Self {
         info!("Dict size in LZ buffer: {}", dict_size);
         Self {
-            stream,
-            buf: vec![0; dict_size],
+            stream: stream,
```
gendx: Clippy suggests to remove the redundant field name.

src/decode/lzbuffer.rs
```diff
@@ -138,13 +138,25 @@ where
...
             len: 0,
         }
     }
+
+    fn get(&self, index: usize) -> u8 {
+        match self.buf.get(index) {
+            Some(x) => *x,
+            None => 0,
+        }
+    }
```
gendx: `self.buf.get(index).unwrap_or(0)`

src/decode/lzbuffer.rs
```diff
@@ -138,13 +138,25 @@ where
...
+
+    fn get_mut(&mut self, index: usize) -> &mut u8 {
+        self.buf.resize(index + 1, 0);
```
gendx: If `index` is less than the current size, this shrinks the buffer, which is incorrect. We probably need more test cases to make sure to capture this.

dragly:
Thanks for the comments and feedback! And sorry about the late reply. I did not have time to get back to this before now. I agree with your input and will make sure we do not risk shrinking the buffer.

src/decode/lzbuffer.rs
dragly: Done, but it needs to be `*self.buf.get(index).unwrap_or(&0)`

dragly:
This took a while, but I have adapted the code to your comments now.

ibaryshnikov:
I can see around 5% regression in speed after this update. Is it possible to use a feature or a runtime option? Or having several buffer implementations, and then passing the one you like from outside?

gendx:
> I can see around 5% regression in speed after this update. Is it possible to use a feature or a runtime option? Or having several buffer implementations, and then passing the one you like from outside?

Good point. I also noticed it in https://github.com/gendx/lzma-rs/pull/22#pullrequestreview-335624836, but I think 5% could be acceptable. However, I've now filed #27 to track this regression, feel free to send a pull-request so that it can be fixed before the next release.

catenacyber:
Just to mention this is a similar intent of what I have done here :
https://github.com/OISF/libhtp/commit/fe16fa764f7cea57be5a288ee85b27dffc460f6f#diff-523a8b9d16e6d9f006c824e158eeb4b7

For resizing, I would suggest it is important to keep track of the dictionary size mentioned in the header, and not go over it.

dragly:
@catenacyber interesting! Thanks for sharing :)

Why is it important not to go over it? Can the decoding fail if we do so, or is there a different reason to stay within the size?

catenacyber:
I am no LZMA expert, so I might be mistaken.

But what is the behavior of other lzma tools, when the header specifies a dictionary size of, for instance, 4096 bytes, and the decompression somehow tries to access item 4097 of dictionary ?
Let me know if I am not clear.
cf line 1041 here : https://github.com/OISF/libhtp/commit/2fb0091981c95a3bf8393703484bcb2871678b63#diff-1143601df653d71da0a4ff7c68de9e29R1041

gendx:
> Just to mention this is a similar intent of what I have done here :
> [OISF/libhtp@fe16fa7#diff-523a8b9d16e6d9f006c824e158eeb4b7](https://github.com/OISF/libhtp/commit/fe16fa764f7cea57be5a288ee85b27dffc460f6f#diff-523a8b9d16e6d9f006c824e158eeb4b7)
> 
> For resizing, I would suggest it is important to keep track of the dictionary size mentioned in the header, and not go over it.

From what I see, the indices passed to `get` are currently always modulo the dict size, so we shouldn't reach past the dictionary size of the header. That said:
- We could add checks in `debug_assertions` mode inside of `get` so that indeed this doesn't happen.
- We should have unit tests for it. Implementing #24 would greatly help to write these kinds of unit tests (and such tests could be helpful to compare to other LZMA parsers as well).

Issue #11, opened 2019-09-09
LZMAError("Expected unpacked size of 149198 but decompressed to 483334")'


David-OConnor:
Any ideas on what could cause this? Code:

```rust
    let mut f = io::BufReader::new(fs::File::open(archive_path).unwrap());
    let mut tar: Vec<u8> = Vec::new();
    lzma_rs::xz_decompress(&mut f, &mut tar).unwrap();
```

gendx:
An LZMA stream can include an unpacked_size hint in its header (https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma.rs#L61-L74), which the code then verifies to reject inconsistencies (https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma.rs#L312-L320).

Additionally, the LZMA2 format is a wrapper around LZMA, which can also provide an unpacked size hint on top of it (https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma2.rs#L89-L95 and https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma2.rs#L161).

On top of that, XZ compresses each file with an LZMA2 stream.

So it looks like either your file was corrupted or there is a bug in my code due to a corner case that I didn't see before.

- Can you comment out the error check (https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma.rs#L312-L320) and let me know if decompression works for your file?
- Do you know which software created this archive?
- Can you run your code with an environment variable set to `RUST_LOG=lzma-rs=info`, so that I can get a clearer idea of what is going on?
- If the file is publicly available (or if you can reproduce the issue on a publicly available file), can you point it to me so that I can debug further?

gendx:
Would https://github.com/gendx/lzma-rs/pull/17 (or a variant of it) work for this use case?

ibaryshnikov:
@gendx I created a reproduction, will it help?
```rust
use std::io::BufReader;
use lzma_rs::decompress::{Options, UnpackedSize};

const DATA: &[u8] = &[
    93, 0, 0, 1, 0, 0, 0, 111, 253, 255, 255, 163, 183, 255, 71, 62, 72, 21, 114, 57, 97, 81, 184,
    146, 40, 230, 143, 221, 66, 251, 179, 253, 113, 133, 36, 209, 157, 136, 6, 166, 184, 144, 144,
[...]

PR #26, opened 2020-01-28
don't check for EOF when unpacked size is specified

ibaryshnikov:
### Pull Request Overview

This pull request fixes the case where the last byte is 0 with known unpacked size.
Related issue is #11, in particular https://github.com/gendx/lzma-rs/issues/11#issuecomment-574590968

### Testing Strategy

The best reference I was able to find is
https://svn.python.org/projects/external/xz-5.0.3/doc/lzma-file-format.txt

```
Uncompressed Size is stored as unsigned 64-bit little endian
integer. A special value of 0xFFFF_FFFF_FFFF_FFFF indicates
that Uncompressed Size is unknown. End of Payload Marker (*)
is used if and only if Uncompressed Size is unknown.
```

### TODO, help wanted

I'm unsure what self.rep[0] is checking here
```rust
if self.rep[0] == 0xFFFF_FFFF {
    if rangecoder.is_finished_ok()? {
        break;
    }
    return Err(error::Error::LZMAError(String::from(
        "Found end-of-stream marker but more bytes are available",
    )));
}
```


dragly:
This fix works for us as well. I agree with what @gendx said in https://github.com/gendx/lzma-rs/issues/11#issuecomment-578167511 and would prefer to find a test-case that reproduces the issue with a proper unpacked size. However, it would be nice if this PR could be merged and an issue created for the test case, if the code is otherwise sound.

ibaryshnikov:
@dragly 

> would prefer to find a test-case that reproduces the issue with a proper unpacked size

I think the test case I submitted has a proper unpacked size. If it doesn't, I'd be curios to know why

dragly:
Sorry, I misread your example. I thought you provided the unpacked data and that the length you gave was different from the actual amount of data.

One option is to just change the `round_trip_file` test to also do a round-trip with unpacked size written to the header (the default is not to write it to the header):

```rust
fn round_trip_file(filename: &str) {
    use std::io::Read;

    let mut x = Vec::new();
    std::fs::File::open(filename)
        .unwrap()
        .read_to_end(&mut x)
        .unwrap();

    round_trip(x.as_slice());

    // Do another round trip, but this time also write it to the header
    let encode_options = lzma_rs::compress::Options {
        unpacked_size: lzma_rs::compress::UnpackedSize::WriteToHeader(Some(x.len() as u64)),
    };
    let decode_options = lzma_rs::decompress::Options {
        unpacked_size: lzma_rs::decompress::UnpackedSize::ReadFromHeader,
    };
    round_trip_with_options(x.as_slice(), &encode_options, &decode_options);
}
```

This actually trips up the `range-coder-edge-case` file with the following error:

```
called `Result::unwrap()` on an `Err` value: LZMAError("Expected unpacked size of 3040092 but decompressed to 3040057")
```

So I think that it is sufficient to modify the test to include the additional round-trip.

However, if you prefer to use your own data (where does it come from, by the way?), you could add the result of uncompressing your bytes as a new `unpacked-size-edge-case` file and change the `round_trip_files` test to

```rust
#[test]
fn round_trip_files() {
    let _ = env_logger::try_init();
    round_trip_file("tests/files/foo.txt");
    round_trip_file("tests/files/range-coder-edge-case");
    round_trip_file("tests/files/unpacked-size-edge-case");
}
```

Either way, it becomes clear that the issue is inherent to the `lzma-rs` library, because it shows that it is not able to properly decompress a file it has itself compressed.

src/decode/lzma.rs
```diff
@@ -225,7 +225,11 @@ where
         rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
     ) -> error::Result<()> {
         loop {
-            if self.unpacked_size.is_some() && rangecoder.is_finished_ok()? {
+            if let Some(unpacked_size) = self.unpacked_size {
+                if self.output.len() as u64 >= unpacked_size {
```
gendx: I'm wondering why the check isn't for equality. A larger output should trigger an error at this point instead.

ibaryshnikov:
@dragly the data is a part of openctm file. It appears that I'm working with wasm, too.

@dragly @gendx I like the idea of using roundrip test instead, I will send an update shortly.

src/decode/lzma.rs
ibaryshnikov: We're just doing `break` here. If it's an equality and the length goes above the expected value, it may take quite some time before the loop ends. With `>=` the program would break immediately and there's still a check later:
```rust
if let Some(len) = self.unpacked_size {
    if self.output.len() as u64 != len {
        return Err(error::Error::LZMAError(format!(
            "Expected unpacked size of {} but decompressed to {}",
            len,
            self.output.len()
        )));
    }
}
```
Is it correct or am I missing something?

gendx: Makes sense given the check right after the loop.

PR #36, opened 2020-04-16
Add decoder fuzzing

Shnatsel:
Version of #34 as a single commit and without the fuzzing corpus

### DIFF
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -1,0 +1,18 @@
+This directory contains fuzzing targets to verify implementation correctness: 
+
+- `roundtrip_*` targets check that we can successfully decode what we've encoded.
+- `decompress_*` targets check that we don't panic or abort on decoding a crafted file.
+- `compare_*` targets check that we produce identical output to liblzma on decompression.
+
+The command to run fuzzer is:
+
+`cargo +nightly fuzz run --release -s none <fuzzing_target>`
+
+For example,
+
+`cargo +nightly fuzz run --release -s none compare_xz`
+
+We use `-s none` because this crate does not contain unsafe code, so we don't
+need sanitizers to detect memory or concurrency errors for us.
+
+For more info see `cargo +nightly fuzz help`
--- a/fuzz/fuzz_targets/compare_xz.rs
+++ b/fuzz/fuzz_targets/compare_xz.rs
@@ -31,6 +31,6 @@ fuzz_target!(|data: &[u8]| {
         (Err(_), Err(_)) => (), // both failed, so behavior matches
         (Ok(_), Err(_)) => panic!("lzma-rs succeeded but xz2 failed"),
         (Err(_), Ok(_)) => panic!("xz2 succeeded but lzma-rs failed"),
-        (Ok(a), Ok(b)) => assert!(a.as_slice() == b.as_slice())
+        (Ok(a), Ok(b)) => assert!(a == b)
     }
 });
--- a/fuzz/fuzz_targets/compare_xz.rs
+++ b/fuzz/fuzz_targets/compare_xz.rs
@@ -4,8 +4,8 @@ extern crate libfuzzer_sys;
 extern crate lzma_rs;
 extern crate xz2;
 
-use std::io::Read;
 use lzma_rs::error::Result;
+use std::io::Read;
 use xz2::stream;
 
 fn decode_xz_lzmars(compressed: &[u8]) -> Result<Vec<u8>> {
@@ -19,7 +19,9 @@ fn decode_xz_xz2(compressed: &[u8]) -> Result<Vec<u8>> {
     let bf = std::io::Cursor::new(compressed);
     let mut decomp: Vec<u8> = Vec::new();
     // create new XZ decompression stream with 8Gb memory limit and checksum verification disabled
-    let xz_stream = stream::Stream::new_stream_decoder(8*1024*1024*1024, stream::IGNORE_CHECK).expect("Failed to create stream");
+    let xz_stream =
+        stream::Stream::new_stream_decoder(8 * 1024 * 1024 * 1024, stream::IGNORE_CHECK)
+            .expect("Failed to create stream");
     xz2::bufread::XzDecoder::new_stream(bf, xz_stream).read_to_end(&mut decomp)?;
     Ok(decomp)
 }
@@ -31,6 +33,6 @@ fuzz_target!(|data: &[u8]| {
         (Err(_), Err(_)) => (), // both failed, so behavior matches
         (Ok(_), Err(_)) => panic!("lzma-rs succeeded but xz2 failed"),
         (Err(_), Ok(_)) => panic!("xz2 succeeded but lzma-rs failed"),
-        (Ok(a), Ok(b)) => assert!(a == b)
+        (Ok(a), Ok(b)) => assert!(a == b),
     }
 });
--- a/fuzz/fuzz_targets/compare_xz.rs
+++ b/fuzz/fuzz_targets/compare_xz.rs
@@ -1,0 +1,36 @@
+#![no_main]
+#[macro_use]
+extern crate libfuzzer_sys;
+extern crate lzma_rs;
+extern crate xz2;
+
+use std::io::Read;
+use lzma_rs::error::Result;
+use xz2::stream;
+
+fn decode_xz_lzmars(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut bf = std::io::Cursor::new(compressed);
+    let mut decomp: Vec<u8> = Vec::new();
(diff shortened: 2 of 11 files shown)


fuzz/README.md
```diff
@@ -0,0 +1,18 @@
...
+The command to run fuzzer is:
+
+`cargo +nightly fuzz run --release -s none <fuzzing_target>`
+
+For example,
+
+`cargo +nightly fuzz run --release -s none compare_xz`
+
+We use `-s none` because this crate does not contain unsafe code, so we don't
+need sanitizers to detect memory or concurrency errors for us.
```
gendx, line 16: In the compare case, shouldn't we sanitize the liblzma part? Or would the sanitizing be scoped to the Rust binding wrapper only, which is likely not so interesting, but not applied to the liblzma itself (being compiled independently by a C/C++ compiler)?

fuzz/fuzz_targets/decompress_lzma.rs
```diff
@@ -0,0 +1,18 @@
...
+fn decode_lzma(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut bf = std::io::Cursor::new(compressed);
+
+    let mut decomp: Vec<u8> = Vec::new();
+    lzma_rs::lzma_decompress(&mut bf, &mut decomp)?;
+    Ok(decomp)
+}
+
+fuzz_target!(|data: &[u8]| {
+    let _decomp = decode_lzma(data);
```
gendx, line 17: When writing benchmarks, I recall the pattern of having a lambda function returning a value of some arbitrary type `T`. Then, the benchmark framework would mark this value as "consumed" to avoid optimizing it away.

Although the goals are different, could a similar pattern be used here? Or does `fuzz_target` require a function returning `()`?

```rust
fuzz_target!(|data: &[u8]| {
    decode_lzma(data)
}
```

fuzz/fuzz_targets/decompress_lzma.rs
```diff
@@ -0,0 +1,18 @@
+#![no_main]
+#[macro_use]
+extern crate libfuzzer_sys;
+extern crate lzma_rs;
+
+use lzma_rs::error::Result;
+
+fn decode_lzma(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut bf = std::io::Cursor::new(compressed);
```
gendx, line 9: In tests, I've been using `std::io::BufReader` around a slice. Which one of a `Cursor` or `BufReader` is more appropriate? The seeking property of `Cursor` is not needed, but is the `BufRead` implementation more efficient for a `Cursor<&[u8]>` or for a `BufReader<&[u8]>` (or are they the same)?

fuzz/fuzz_targets/compare_xz.rs
```diff
@@ -0,0 +1,36 @@
...
+    let mut decomp: Vec<u8> = Vec::new();
+    lzma_rs::xz_decompress(&mut bf, &mut decomp)?;
+    Ok(decomp)
+}
+
+fn decode_xz_xz2(compressed: &[u8]) -> Result<Vec<u8>> {
+    let bf = std::io::Cursor::new(compressed);
+    let mut decomp: Vec<u8> = Vec::new();
+    // create new XZ decompression stream with 8Gb memory limit and checksum verification disabled
+    let xz_stream = stream::Stream::new_stream_decoder(8589935000, stream::IGNORE_CHECK).expect("Failed to create stream");
```
gendx, line 22: Where does the exact value `8589935000` come from? Is it better represented in hexadecimal form? Is it a magic value provided by the `xz2` crate, or by liblzma?

fuzz/fuzz_targets/compare_xz.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+fuzz_target!(|data: &[u8]| {
+    let result_lzmars = decode_xz_lzmars(data);
+    let result_xz2 = decode_xz_xz2(data);
+    match (result_lzmars, result_xz2) {
+        (Err(_), Err(_)) => (), // both failed, so behavior matches
+        (Ok(_), Err(_)) => panic!("lzma-rs succeeded but xz2 failed"),
+        (Err(_), Ok(_)) => panic!("xz2 succeeded but lzma-rs failed"),
+        (Ok(a), Ok(b)) => assert!(a.as_slice() == b.as_slice())
```
gendx, line 34: I see:

```rust
impl<A, B> PartialEq<Vec<B>> for Vec<A>
where
    A: PartialEq<B>
{ ... }

impl<T: Eq> Eq for Vec<T> {}
```

so the `as_slice` part shouldn't be necessary.

fuzz/README.md
Shnatsel: > Or would the sanitizing be scoped to the Rust binding wrapper only, which is likely not so interesting, but not applied to the liblzma itself 

Yes, that's the case.

Besides, I'm not particularly interested in fuzzing liblzma itself because [Google already fuzzes it](https://bugs.chromium.org/p/oss-fuzz/issues/list?q=-status%3AWontFix%2CDuplicate%20-component%3AInfra%20proj%3Dlzma&can=1).

fuzz/fuzz_targets/decompress_lzma.rs
Shnatsel: Yes, returning `()` is required by the function signature.

fuzz/fuzz_targets/decompress_lzma.rs
Shnatsel: Swapping one for the other didn't result in any measurable difference in fuzzing speed.

fuzz/fuzz_targets/compare_xz.rs
Shnatsel: That's 8 gibibytes in bytes. Frankly the value is quite arbitrary, I just wanted to have something as a starting point.

Shnatsel: I've rewritten it in a more readable way

Shnatsel:
It has occurred to me that completely identical error behaviour on arbitrary input is a pretty high bar, so it would be nice to have a less stringent test. I've added fuzzing targets checking that data compressed by the reference implementation can be decompressed by lzma-rs and vice versa. They seem to pass.

fuzz/fuzz_targets/compare_xz.rs
```diff
@@ -0,0 +1,36 @@
...
+    let mut decomp: Vec<u8> = Vec::new();
+    lzma_rs::xz_decompress(&mut bf, &mut decomp)?;
+    Ok(decomp)
+}
+
+fn decode_xz_xz2(compressed: &[u8]) -> Result<Vec<u8>> {
+    let bf = std::io::Cursor::new(compressed);
+    let mut decomp: Vec<u8> = Vec::new();
+    // create new XZ decompression stream with 8Gb memory limit and checksum verification disabled
+    let xz_stream = stream::Stream::new_stream_decoder(8*1024*1024*1024, stream::IGNORE_CHECK).expect("Failed to create stream");
```
gendx, line 22: I'm surprised that the `cargo fmt -- --check` didn't trigger.
I'm even more surprised that Travis-CI didn't report any test results on the pull request.

fuzz/fuzz_targets/compare_xz.rs
gendx: Well, `8*1024*1024*1024` is `8589934592`, not `8589935000` so that surprised me at first. I think `0x2_0000_0000` would be more readable (`0x2` followed by 8 nibbles of zeros = 32 bits of zeros).

gendx:
> It has occurred to me that completely identical error behaviour on arbitrary input is a pretty high bar, so it would be nice to have a less stringent test. I've added fuzzing targets checking that data compressed by the reference implementation can be decompressed by lzma-rs and vice versa. They seem to pass.

These checks are definitely useful in addition!
- Decoding random inputs exercises failure checks.
- Decoding well-encoded inputs exercises deeper decoding paths.

My last comments are mostly nit picking at this point. In addition, if you (or anyone reading this) happen to have an idea about why my Travis-CI/Bors integration is broken, that would be very useful.

Shnatsel:
I see the PR has "waiting on author" tag, but it's not clear to me what still needs to be changed. Could you provide me with a list of what needs to be done before this can be merged?

gendx:
> I see the PR has "waiting on author" tag, but it's not clear to me what still needs to be changed. Could you provide me with a list of what needs to be done before this can be merged?

The following comment is still pending (in particular, running `cargo fmt`): https://github.com/gendx/lzma-rs/pull/36#discussion_r411203572

Shnatsel:
I've run `cargo fmt` and migrated the fuzzing harness to 2018 edition. If there are any other blockers, please let me know.

Issue #28, opened 2020-03-05
Travis/Bors integration broken


gendx:
Bors timed out in the last two pull-requests (#22 and #26). The `continuous-integration/travis-ci/pr` doesn't show up on the relevant commits. However, some jobs appear in [Travis build history](https://travis-ci.org/gendx/lzma-rs/builds) (`staging` branch).

I've had a quick look at the [bors docs](https://bors.tech/documentation/getting-started/), but didn't find an obvious mis-configuration.

If anyone has an idea about why Bors could fail, help is welcome!

gendx:
Travis is long gone, and Bors isn't really useful anyway now that GitHub Actions are fully-fledged.

PR #38, opened 2020-04-23
lib: remove spurious extern statements

lucab:
This removes `extern crate` statements, not required on 2018 edition.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,11 +1,5 @@
 #![forbid(unsafe_code)]
 
-extern crate byteorder;
-#[cfg(feature = "enable_logging")]
-#[macro_use]
-extern crate log;
-extern crate crc;
-
 #[macro_use]
 mod macros;
 
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2,7 +2,7 @@
 #[macro_export]
 macro_rules! lzma_trace {
     ($($arg:tt)+) => {
-        trace!($($arg)+);
+        log::trace!($($arg)+);
     }
 }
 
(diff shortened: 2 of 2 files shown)


lucab:
@gendx related, what is the Minimum Supported Rust Version (MSRV) targeted by this crate?

lucab:
Not sure why Travis did not report back (missing repo permissions?), but CI is green: https://travis-ci.org/github/gendx/lzma-rs/builds/678510618.

gendx:
> @gendx related, what is the Minimum Supported Rust Version (MSRV) targeted by this crate?

For now there's none specified. The Travis-CI config uses whatever the current `stable` is (https://github.com/gendx/lzma-rs/blob/master/.travis.yml).

Any pull request to add a specific version to Travis-CI and mark it as the MSRV in the README will be appreciated!

---

Related, I'm still wondering why Travis status isn't shown here. But this build succeeded: https://travis-ci.org/github/gendx/lzma-rs/builds/678510618. Any help to fix #28 is appreciated.

lucab:
@gendx do you have any target version in mind? Maybe as a start, current stable minus two, i.e. 1.40? Anything >= 1.32 would also help in dropping the `byteorder` dependency.

gendx:
I was thinking about some low version for now (1.32 seems reasonable) - to avoid potential compatibility issues with the 2 dependent crates: https://crates.io/crates/lzma-rs/reverse_dependencies
In principle, `lzma-rs` doesn't depend on many things, so I expect a low version to work.

I'm also planning to release a 0.1.3 version soon to include some recent performance improvements and bug fixes. So not breaking the dependents would be a nice goal for that.

Then, we can bump up the MSRV for a 0.2.x version if that unlocks more possibilities.

Issue #32, opened 2020-04-15
Panic on decoding xz file with sha256 checksum


Shnatsel:
https://github.com/gendx/lzma-rs/blob/3fcf0bbfc5f13f22a35d397c568387ec4cfac8e4/src/decode/xz.rs#L364

This panic is trivial to trigger and may be used to mount a denial-of-service attack on a remote service that decodes XZ files.

Returning an error instead of panicking would be an improvement. https://crates.io/crates/sha2 can be used to implement the check.

Issue #6, opened 2017-10-17
Improve error interface


gendx:
Use various enums instead of the current `LZMAError(String)`.

PR #40, opened 2020-04-24
decode/xz: do not panic on unsupported checksum

lucab:
This returns an error when trying to decode inputs with an unsupported
SHA-256 checksum, instead of panicking.

Closes: https://github.com/gendx/lzma-rs/issues/32

### DIFF
--- a/src/decode/xz.rs
+++ b/src/decode/xz.rs
@@ -361,7 +361,11 @@ where
             }
         }
         // TODO
-        CheckMethod::SHA256 => unimplemented!(),
+        CheckMethod::SHA256 => {
+            return Err(error::Error::XZError(
+                "Unsupported SHA-256 checksum (not yet implemented)".to_string(),
+            ));
+        }
     }
     Ok(())
 }


src/decode/xz.rs
```diff
@@ -361,7 +361,11 @@ where
             }
         }
         // TODO
-        CheckMethod::SHA256 => unimplemented!(),
+        CheckMethod::SHA256 => {
+            return Err(error::Error::XZError(
+                "unsupported SHA-256 checksum (not yet implemented)".to_string(),
```
gendx, line 366: Nit: the other error messages start with an uppercase letter.

lucab: Ah, good catch, aligned.
Though I usually see libraries not uppercasing, as these messages are usually wrapped by applications and printed like `app logic failed: library error details`, FWIW.

gendx: The error interface should be improved anyway. See #6.

gendx:
Waiting a bit for Travis.

PR #41, opened 2020-04-25
xz: unify some common logic

lucab:
This reworks and unifies some logic which is shared between XZ encoder
and decoder.

### DIFF
--- a/src/decode/xz.rs
+++ b/src/decode/xz.rs
@@ -3,8 +3,8 @@
 use crate::decode::lzma2;
 use crate::decode::util;
 use crate::error;
-use crate::xz::{footer, header};
-use byteorder::{LittleEndian, ReadBytesExt};
+use crate::xz::{footer, header, CheckMethod, StreamFlags};
+use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
 use crc::{crc32, crc64, Hasher32};
 use std::hash::Hasher;
 use std::io;
@@ -21,11 +21,7 @@ where
     R: io::BufRead,
     W: io::Write,
 {
-    let header = {
-        let mut header_buf: [u8; 12] = [0; 12];
-        input.read_exact(&mut header_buf)?;
-        header::StreamHeader::parse(&header_buf)?
-    };
+    let header = header::StreamHeader::parse(input)?;
 
     let mut records: Vec<Record> = vec![];
     let index_size = loop {
@@ -43,7 +39,7 @@ where
         read_block(
             &mut count_input,
             output,
-            header.check_method,
+            header.stream_flags.check_method,
             &mut records,
             header_size,
         )?;
@@ -64,19 +60,15 @@ where
             )));
         }
 
-        let null_byte = digested.read_u8()?;
-        if null_byte != 0x00 {
-            return Err(error::Error::XZError(format!(
-                "Invalid null byte in footer: {:x}",
-                null_byte
-            )));
-        }
+        let stream_flags = {
+            let field = digested.read_u16::<BigEndian>()?;
+            StreamFlags::parse(field)?
+        };
 
-        let check_method = digested.read_u8()?;
-        if (header.check_method as u8) != check_method {
+        if header.stream_flags != stream_flags {
             return Err(error::Error::XZError(format!(
                 "Flags in header ({:?}) does not match footer ({:?})",
-                header.check_method, check_method
+                header.stream_flags, stream_flags
             )));
         }
     }
@@ -209,7 +201,7 @@ struct BlockHeader {
 fn read_block<'a, R, W>(
     count_input: &mut util::CountBufRead<'a, R>,
     output: &mut W,
-    check_method: header::CheckMethod,
+    check_method: CheckMethod,
     records: &mut Vec<Record>,
     header_size: u8,
 ) -> error::Result<bool>
@@ -301,17 +293,13 @@ where
     Ok(finished)
 }
 
-fn check_checksum<R>(
-    input: &mut R,
-    buf: &[u8],
-    check_method: header::CheckMethod,
-) -> error::Result<()>
+fn check_checksum<R>(input: &mut R, buf: &[u8], check_method: CheckMethod) -> error::Result<()>
 where
     R: io::BufRead,
 {
     match check_method {
-        header::CheckMethod::None => (),
-        header::CheckMethod::CRC32 => {
+        CheckMethod::None => (),
+        CheckMethod::CRC32 => {
             util::discard(input, 4)?;
             let crc32 = input.read_u32::<LittleEndian>()?;
             let digest_crc32 = crc32::checksum_ieee(buf);
@@ -322,7 +310,7 @@ where
                 )));
             }
         }
-        header::CheckMethod::CRC64 => {
+        CheckMethod::CRC64 => {
             let crc64 = input.read_u64::<LittleEndian>()?;
             let digest_crc64 = crc64::checksum_ecma(buf);
             if crc64 != digest_crc64 {
@@ -333,7 +321,7 @@ where
             }
         }
         // TODO
-        header::CheckMethod::SHA256 => {
+        CheckMethod::SHA256 => {
             return Err(error::Error::XZError(
                 "Unsupported SHA-256 checksum (not yet implemented)".to_string(),
             ));
--- a/src/decode/xz.rs
+++ b/src/decode/xz.rs
@@ -1,35 +1,14 @@
+//! Decoder for the `.xz` file format.
+
 use crate::decode::lzma2;
 use crate::decode::util;
 use crate::error;
-use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
+use crate::xz::{footer, header};
+use byteorder::{LittleEndian, ReadBytesExt};
 use crc::{crc32, crc64, Hasher32};
 use std::hash::Hasher;
 use std::io;
 use std::io::Read;
-
-const XZ_MAGIC: &[u8] = &[0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00];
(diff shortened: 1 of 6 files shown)


src/xz/header.rs
```diff
@@ -0,0 +1,111 @@
...
+                magic
+            )));
+        }
+
+        let digested = {
+            let mut digest = crc32::Digest::new(crc32::IEEE);
+            digest.write_u8(null_byte);
+            digest.write_u8(check_id);
+            digest.sum32()
+        };
```
gendx, line 39: Is there a reason to change the logic from a `input: Read` to using a temporary buffer?
- This is in principle less efficient.
- Everything in `lzma-rs` has APIs over generic `io::(Buf)Read` and `io::(Buf)Write`.
- This has nothing to do with this PR (unifying some common logic).

src/xz/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Logic for handling `.xz` file format.
+//!
+//! Format specifications are at [https://tukaani.org/xz/xz-file-format.txt](spec).
+//!
+//! [spec]: https://tukaani.org/xz/xz-file-format.txt
+
+#![deny(missing_docs)]
+#![deny(missing_debug_implementations)]
```
gendx, line 8: Thanks! :)

src/encode/xz.rs
```diff
@@ -28,25 +24,30 @@ where
...
 where
     W: io::Write,
 {
-    output.write_all(XZ_MAGIC)?;
+    output.write_all(header::XZ_MAGIC)?;
     let mut digest = crc32::Digest::new(crc32::IEEE);
     {
         let mut digested = util::HasherWrite::new(output, &mut digest);
-        digested.write_u16::<BigEndian>(flags)?;
+        digested.write_u8(0)?;
```
gendx, line 38: This zero seems weird here. I think we should keep some `Flags` abstraction, which can be created from a `CheckMethod` (for now we don't use other flags), and converted into a `u16` (we may want to add other flags in the future).

src/decode/xz.rs
```diff
@@ -104,11 +64,19 @@ where
             )));
         }
 
-        let footer_flags = digested.read_u16::<BigEndian>()?;
-        if flags != footer_flags {
```
gendx, line 108: We should keep the `u16` flags abstraction, and then check that the flags are valid. Otherwise `Invalid null byte in footer` doesn't explain where this null byte comes form, which is less clear than the code before this pull request.

src/xz/header.rs
```diff
@@ -0,0 +1,111 @@
...
+/// This does not store the null byte in Stream Flags, which is currently unused.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct StreamHeader {
+    pub(crate) check_method: CheckMethod,
+}
+
+impl StreamHeader {
+    /// Parse a Stream Header from a byte buffer.
+    pub(crate) fn parse(input: &[u8; 12]) -> error::Result<Self> {
+        use std::hash::Hasher;
```
gendx, line 20: I didn't check if there is a clippy lint to forbid this, but the style in `lzma-rs` is to import symbols at a module/file level.

lucab: I'm not aware of any such clippy lints. I'll move this out.

src/xz/header.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_checkmethod_roundtrip() {
+        let mut valid = 0;
```
gendx, line 101: Nit: `count_valid` would be more descriptive.

src/xz/header.rs
lucab: I split the buffer reading out of content parsing for the following reasons:
 * there are some checks which have to be performed in reverse order (e.g. CRC check before parsing the preceding Flags value)
 * the header is fixed size, so it can be read in one shot and then parsed (i.e. short reads are direct errors, no need to start parsing)
 * function signature enforces the actual required input size

You are right this goes a bit further than just moving code around, but I got caught into this when splitting the Flags/CheckMethod parsers.

Also (I may be wrong here) I was under the impression that reading the whole header content in one shot would result in better I/O profile than pulling it byte-by-byte.

gendx: In terms of performance:
- Avoiding pulling bytes one-by-one at the I/O layer is the reason why the actual trait bound is `io::BufRead` (I incorrectly mentioned `io::Read`). This already buffers external I/O.
- `u32::from_le_bytes([input[8], input[9], input[10], input[11]])` is more reading byte-by-byte than `input.read_u32::<LittleEndian>()?;` (although the compiler may optimize that away, on common CPU architectures at least).

In any case, performance of reading the header is likely insignificant. What matters is code readability.

I think that cherry-picking the bytes one-by-one is less readable than repeated calls to `read`. In particular, repeated `read`s are guaranteed to be sequential, whereas indexing the bytes may lead to errors due to leaving a gap or reusing a byte twice. So it's clearer in terms of parsing structure (linear parsing, rather than recursive parsing with the "one shot" read of 12 bytes, followed by potentially unstructured reads within the read).

IMO, reading to a buffer to then convert that into some struct is a reasonable pattern to use in C, to obtain good performance despite the lack of high-level abstraction in the language. But in Rust, repeated reads on a `io::BufRead` will likely yield the same (or better) performance, because Rust's generics, despite exposing a high-level API, will be optimized by the compiler under the hood.

In any case, `lzma-rs` supports `cargo bench`, so you can run the benchmarks (and add your own) to see if something increases performance. But unless some benchmark shows an improvement, I don't think it makes sense to make some code change because someone has an intuition that performance would improve. And I'd prefer to keep performance-related pull-requests separate form other pull-requests (code cleanup, refactoring, new features), so that it's easier to review them. Pull-requests that mix multiple things are generally harder to review.

---

Regarding checking the CRC, I don't think it really matters if we do it before or after parsing the header. I don't even think it's useful to have such a CRC in the middle of the file format, for the following reasons.
- Integrity should be the responsibility of the transport layer (e.g. transferring a file over HTTPS) or storage layer (e.g. error-correcting code in a disk).
- There is a magic signature at the beginning of the file to guard against "I tried to decode the wrong file by mistake (`unxz foo.jpg`)".
- CRC is not cryptographically secure, so even unrelated data could "by chance" have the correct CRC.
- If someone can tamper with a file, they can as well put a correct checksum.
- It makes it harder to fuzz.
- It's taking space. A bit ironic for a compression format.

The only use case I see for a CRC is to have a "checkpoint" when someone is writing a parser from scratch. One doesn't have a full parser yet, but can confirm that the header parses correctly, yay! But beyond this development use case I don't see much use of the CRC in production. In general this "I'm writing a parser and want early feedback" problem would be better solved by a careful test suite (with a progression from simple tests that are easy to pass to more complete tests). And simpler specs are also more straightforward to implement without checkpoints - but the checksums themselves make the specs more complex.

Of course, once the CRC exists, checking it is useful for interoperability, e.g. if decoding software X is checking the CRC, we want to obtain the same error result as them when the CRC is wrong (regardless of whether the rest of the file would parse). But maybe software Y is ignoring the checksums, so we could add a decoding option to `lzma-rs` to ignore all (or some) internal checksums, in order to be interoperable with Y.

And that's again one point against having a CRC in such a file format: it creates a decision point (whether to check the CRC or not), which can split the ecosystem in two (decoders that enforce the CRC vs. those that ignore it) times two (encoders that generate a correct CRC vs. those that generate an invalid/blank one). And another ecosystem split point is software that don't support some checksum algorithm (yet), e.g. SHA256 for `lzma-rs` as of today. Supporting an additional checksum algorithm is more code after all, and more code means: more development time, more dependencies to rely on, larger binary size.

My 2 cents on checksums within file formats.

lucab: Thanks for taking the time to write long-form feedback, much appreciated!

I see your rationale and preference for linear parsing out of a bufreader and it totally makes sense, I'll switch back to that.

Regarding the CRC, the last part of [spec section 2.1.1.3](https://tukaani.org/xz/xz-file-format.txt) goes a bit more into details about its role here, in order to tell apart bitflips from new features using reserved bits.

gendx: > This way old decoders will be able to verify the CRC32 calculated from Stream Flags, and thus distinguish between corrupt files (CRC32 doesn't match) and files that the decoder doesn't support (CRC32 matches but Stream Flags has reserved bits set).

I'm still not convinced that the xz layer should have to detect corrupt files (as opposed to integrity in the transport/network layer and error correction in the file system), but well the spec is the spec.

src/xz/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        let flags = Self {
+            check_method: CheckMethod::try_from(flags_bytes[1])?,
+        };
+        Ok(flags)
+    }
+
+    /// Serialize Stream Flags into a writer.
+    pub(crate) fn serialize<W>(self, writer: &mut W) -> io::Result<usize>
```
lucab, line 43: @gendx I was unsure whether you wanted a `to_u16` here, or this kind of directly writing method.

gendx: It's fine within the `StreamFlags` abstraction.

gendx:
Waiting a bit for the CI build to complete: https://travis-ci.org/github/gendx/lzma-rs/builds/680622327
Although the current backlog of MacOS builds is quite high... https://www.traviscistatus.com/#week

lucab:
Should I squash this or do you plan to that yourself when merging?

lucab:
Travis completed, all green: https://travis-ci.org/github/gendx/lzma-rs/builds/680622327

gendx:
> Should I squash this or do you plan to do that yourself when merging?

Not really necessary to squash for such small changes IMO.

PR #44, opened 2020-05-02
lib: add missing docstrings

lucab:
This adds all missing docstrings on public items, and enforces
linting on them.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,3 +1,4 @@
+/// Log trace message (feature: enabled).
 #[cfg(feature = "enable_logging")]
 #[macro_export]
 macro_rules! lzma_trace {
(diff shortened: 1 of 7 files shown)


lucab:
CI is green: https://travis-ci.org/github/gendx/lzma-rs/builds/682309490

gendx:
Really appreciated, thanks! :)

src/macros.rs
```diff
@@ -1,3 +1,4 @@
+/// Log trace message (feature: enabled).
```
lucab, line 1: @gendx as a sidenote, these macros end up in the public API, but I'm not sure whether you really intended that. If you are planning a breaking release soon, you may consider taking them out.

gendx: Indeed. The public API should just consist of the `enable_logging` feature, these macros should only be useful internally (and indeed I don't want any dependency to rely on them, which would add maintenance burden).

PR #50, opened 2020-07-09
lzbuffer: add memlimit option

cccs-sadugas:
### Pull Request Overview

Adds a memlimit configuration option for decompression. If the dict
buffer's memory limit is exceeded, decompression will fail with an
LZMAError. Additional functions were added to reduce the amount of
breaking changes in the library.

Implements #49.

### Testing Strategy

This pull request was tested by...

- [x] Added relevant unit tests.
- [ ] Added relevant end-to-end tests (such as `.lzma`, `.lzma2`, `.xz` files).

### Supporting Documentation and References
* This idea comes from a port of the LZMA SDK integrated in libhtp and used by suricata [here](https://github.com/OISF/libhtp/blob/0.5.x/htp/lzma/LzmaDec.c#L638).

Issue #10, opened 2019-08-28
support streaming read


vn971:
Currently,  one blocking function is provided that reads from `io::BufRead` and writes to `io::Write`. This enforces the user of the library to read all contents into memory, or into a file.

Sometimes, however, it is only needed to traverse the data, but not have it all at once.

Such a thing could be achieved by having a function that, given `io::Read`, gives something that implements `io::Read` as well. This way, you can progressively read on compressed OR decompressed stream, while the library will internally read the decompressed or compressed stream. This is how `xz2` crate works, for example, see the function signature of `xz2::read::XzDecoder::new`. This also looks very flexible and intuitive as well: decompressor starts to act like a "pipe" (in unix terminology), rather than something that writes.

Support of it in lzma-rs would be very nice I think.
Personally, I'm raising the issue because I wanted to try this library in rua https://github.com/vn971/rua  Here I am using an intermediate layer of decompression for another function that accepts `Read` https://github.com/vn971/rua/blob/master/src/tar_check.rs#L26 (however, the underlying library `xz2` is not pure Rust, but uses bindings)

Thoughts?

gendx:
This is a very interesting point!

I was wondering whether there was a generic way of transforming an `io::Write` into an `io::Read`. The opposite would be quite simple (read bytes from an `io::Read` and write them into an `io::Write`), but this looks trickier. Maybe that could be possible with async functions/generators? Or with a separate process - or simply a thread - that "writes" data to the main thread, which reads it (like with Unix pipes).

In the meantime, I think the easiest way to support streaming would be to extract the loop body of the `process` function (https://github.com/gendx/lzma-rs/blob/master/src/decode/lzma.rs#L215) into a `step` function. Then, in the streaming case, use a temporary buffer as the `io::Write` for the current decoder ; the `read` method of your `io::Read` would repeatedly call `step` and copy the bytes from the tmp buffer into the read buffer.

I probably won't have time to look at it more closely this week, but feel free to send a PR if you want to give it a try!

vn971:
Regarding the `process` function and the temporary buffer -- indeed this is how I thought it can be done as well.
I'm not sure I'll have time in the coming days as well though. Maybe I'll come to that later if/when I get rid of other libraries that bind to OS libraries, and will be otherwise on pure Rust.

demurgos:
Hi,
[...]

PR #51, opened 2020-07-13
Streaming Decompressor

cccs-sadugas:
### Pull Request Overview

This pull request implements issue #10 .

It is currently dependant on #50 .

### Testing Strategy

This implementation is based off of the [LzmaDec_TryDummy](https://github.com/OISF/libhtp/blob/0.5.x/htp/lzma/LzmaDec.c#L674) function in libhtp's port of the LZMA SDK.

### TODO

Could hide this under a feature flag upon request. 

### Benchmarks

I added a benchmark `decompress_stream_big_file` that shows that streaming decompression is actually after. This could be related better memory management.

CPU: Intel(R) Core(TM) i7-1065G7 @ 1.30 GHz
RAM: 16Gb @ 4267 MHz
OS: Docker Windows Ubuntu 20.04 LTS

**master**
```
test compress_65536                  ... bench:   2,689,440 ns/iter (+/- 667,102)
test compress_empty                  ... bench:       1,629 ns/iter (+/- 346)
test compress_hello                  ... bench:       2,278 ns/iter (+/- 430)
test decompress_after_compress_65536 ... bench:   3,655,315 ns/iter (+/- 6,067,710)
test decompress_after_compress_empty ... bench:       3,977 ns/iter (+/- 1,797)
test decompress_after_compress_hello ... bench:       4,633 ns/iter (+/- 1,811)
test decompress_big_file             ... bench:   7,079,015 ns/iter (+/- 2,228,350)
test decompress_huge_dict            ... bench:       5,359 ns/iter (+/- 4,065)
```
**streaming-decompressor**
```
test compress_65536                  ... bench:   2,746,007 ns/iter (+/- 1,179,775)
test compress_empty                  ... bench:       1,618 ns/iter (+/- 1,428)
test compress_hello                  ... bench:       3,729 ns/iter (+/- 4,189)
test decompress_after_compress_65536 ... bench:   3,134,650 ns/iter (+/- 2,263,524)
test decompress_after_compress_empty ... bench:       3,747 ns/iter (+/- 1,575)
test decompress_after_compress_hello ... bench:       4,558 ns/iter (+/- 1,288)
test decompress_big_file             ... bench:   7,212,875 ns/iter (+/- 2,376,085)
test decompress_huge_dict            ... bench:       4,633 ns/iter (+/- 7,197)
test decompress_stream_big_file      ... bench:   6,670,270 ns/iter (+/- 2,605,148)
```

### DIFF
--- a/benches/lzma.rs
+++ b/benches/lzma.rs
@@ -2,7 +2,7 @@
 
 extern crate test;
 
-use std::io::Read;
+use std::io::{Read, Write};
 use test::Bencher;
 
 fn compress_bench(x: &[u8], b: &mut Bencher) {
@@ -31,11 +31,26 @@ fn decompress_bench(compressed: &[u8], b: &mut Bencher) {
     });
 }
 
+fn decompress_stream_bench(compressed: &[u8], b: &mut Bencher) {
+    b.iter(|| {
+        let mut stream = lzma_rs::decompress::Stream::new(Vec::new());
+        stream.write_all(compressed).unwrap();
+        stream.finish().unwrap();
+    });
+}
+
 fn decompress_bench_file(compfile: &str, b: &mut Bencher) {
     let mut f = std::fs::File::open(compfile).unwrap();
     let mut compressed = Vec::new();
     f.read_to_end(&mut compressed).unwrap();
     decompress_bench(&compressed, b);
+}
+
+fn decompress_stream_bench_file(compfile: &str, b: &mut Bencher) {
+    let mut f = std::fs::File::open(compfile).unwrap();
+    let mut compressed = Vec::new();
+    f.read_to_end(&mut compressed).unwrap();
+    decompress_stream_bench(&compressed, b);
 }
 
 #[bench]
@@ -85,6 +100,13 @@ fn decompress_big_file(b: &mut Bencher) {
     #[cfg(feature = "enable_logging")]
     let _ = env_logger::try_init();
     decompress_bench_file("tests/files/foo.txt.lzma", b);
+}
+
+#[bench]
+fn decompress_stream_big_file(b: &mut Bencher) {
+    #[cfg(feature = "enable_logging")]
+    let _ = env_logger::try_init();
+    decompress_stream_bench_file("tests/files/foo.txt.lzma", b);
 }
 
 #[bench]
--- a/src/decode/lzbuffer.rs
+++ b/src/decode/lzbuffer.rs
@@ -8,7 +8,7 @@ pub trait LZBuffer {
     // Retrieve the n-th last byte
     fn last_n(&self, dist: usize) -> error::Result<u8>;
     // Append a literal
-    fn append_literal(&mut self, lit: u8) -> io::Result<()>;
+    fn append_literal(&mut self, lit: u8) -> error::Result<()>;
     // Fetch an LZ sequence (length, distance) from inside the buffer
     fn append_lz(&mut self, len: usize, dist: usize) -> error::Result<()>;
     // Flush the buffer to the output
@@ -22,6 +22,7 @@ where
 {
     stream: &'a mut W, // Output sink
     buf: Vec<u8>,      // Buffer
+    memlimit: usize,   // Buffer memory limit
     len: usize,        // Total number of bytes sent through the buffer
 }
 
@@ -30,9 +31,14 @@ where
     W: io::Write,
 {
     pub fn from_stream(stream: &'a mut W) -> Self {
+        Self::from_stream_with_memlimit(stream, std::usize::MAX)
+    }
+
+    pub fn from_stream_with_memlimit(stream: &'a mut W, memlimit: usize) -> Self {
         Self {
             stream,
             buf: Vec::new(),
+            memlimit,
             len: 0,
         }
     }
@@ -84,10 +90,19 @@ where
     }
 
     // Append a literal
-    fn append_literal(&mut self, lit: u8) -> io::Result<()> {
-        self.buf.push(lit);
-        self.len += 1;
-        Ok(())
+    fn append_literal(&mut self, lit: u8) -> error::Result<()> {
+        let new_len = self.len + 1;
+
+        if new_len > self.memlimit {
+            Err(error::Error::LZMAError(format!(
+                "exceeded memory limit of {}",
+                self.memlimit
+            )))
+        } else {
+            self.buf.push(lit);
+            self.len = new_len;
+            Ok(())
+        }
     }
 
     // Fetch an LZ sequence (length, distance) from inside the buffer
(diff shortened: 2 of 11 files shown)


src/decode/stream.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// we have properly reached the end of the stream.
+    pub fn finish(mut self) -> Result<W, std::io::Error> {
+        if let Some(state) = self.state.take() {
+            match state {
+                State::Init(output) => {
+                    if self.tmp_len > 0 {
+                        Err(std::io::Error::new(
+                            std::io::ErrorKind::Other,
+                            "failed to read header",
+                        ))
```
gendx, line 128: This should be an `LZMAError`.

src/decode/stream.rs
```diff
@@ -0,0 +1,476 @@
...
+        let params = match LZMAParams::read_header(&mut input, options) {
+            Ok(params) => params,
+            Err(e) => {
+                return Err((Some(State::Init(output)), e.into()));
+            }
+        };
+
+        let len = match input.fill_buf() {
+            Ok(val) => val,
+            Err(_) => {
```
gendx, line 175: The original I/O error shouldn't be lost. It can provide useful information to the user (network error, invalid permissions to read a file, etc.).

src/decode/stream.rs
```diff
@@ -0,0 +1,476 @@
...
+
+    /// Attempts to read the header and transition into a running state.
+    ///
+    /// This function will consume the state, returning the next state on both
+    /// error and success.
+    fn read_header<R: BufRead>(
+        output: W,
+        mut input: &mut R,
+        options: &Options,
+    ) -> Result<State<W>, (Option<State<W>>, std::io::Error)> {
```
gendx, line 165: `(Option<State<W>>, std::io::Error)` looks weird for an error type. Can you extend `LZMAError` for this use case?

src/decode/stream.rs
```diff
@@ -0,0 +1,476 @@
...
+            assert!(err.to_string().contains("failed to fill whole buffer"));
+
+            end += 1;
+        }
+    }
+
+    /// Test processing all chunk sizes
+    #[test]
+    fn test_stream_chunked() {
+        let small_input = b"Project Gutenberg's Alice's Adventures in Wonderland, by Lewis Carroll";
```
gendx, line 413: This looks like a useful input to have as a test file, so that other unit tests can use it as well.

benches/lzma.rs
```diff
@@ -31,13 +31,28 @@ fn decompress_bench(compressed: &[u8], b: &mut Bencher) {
     });
 }
 
+fn decompress_stream_bench(compressed: &[u8], b: &mut Bencher) {
+    b.iter(|| {
+        let mut stream = lzma_rs::decompress::Stream::new(Vec::new());
+        stream.write_all(compressed).unwrap();
+        stream.finish().unwrap();
```
gendx, line 38: I belive the resulting decompressed output should be the result of the lambda function passed to `Bencher::iter`, to make sure it's not optimized away. However, other benchmarks don't currently do that, but I can send a pull request to fix that accordingly.

cccs-sadugas:
Thanks :). I will address your comments and open a new PR.

tests/stream.rs
```diff
@@ -0,0 +1,273 @@
...
+        ..Default::default()
+    };
+    round_trip_with_options(x, &encode_options, &decode_options);
+}
+
+fn round_trip_no_options(x: &[u8]) {
+    let mut compressed: Vec<u8> = Vec::new();
+    lzma_rs::lzma_compress(&mut std::io::BufReader::new(x), &mut compressed).unwrap();
+    #[cfg(feature = "enable_logging")]
+    info!("Compressed {} -> {} bytes", x.len(), compressed.len());
```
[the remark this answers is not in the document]
cccs-sadugas: I'm not sure those macros can be imported from the tests folder. This code was copied from [`tests/lzma.rs`](https://github.com/gendx/lzma-rs/blob/master/tests/lzma.rs#L21).

src/decode/lzbuffer.rs
```diff
@@ -1,38 +1,51 @@
...
     fn last_n(&self, dist: usize) -> error::Result<u8>;
     // Append a literal
-    fn append_literal(&mut self, lit: u8) -> io::Result<()>;
+    fn append_literal(&mut self, lit: u8) -> error::Result<()>;
     // Fetch an LZ sequence (length, distance) from inside the buffer
     fn append_lz(&mut self, len: usize, dist: usize) -> error::Result<()>;
+    // Get a reference to the output sink
+    fn get_ref(&self) -> &W;
+    // Get a mutable reference to the output sink
+    fn get_mut(&mut self) -> &mut W;
```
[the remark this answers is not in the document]
cccs-sadugas: I went with `get_output` and `get_output_mut`.

src/decode/rangecoder.rs
```diff
@@ -92,6 +134,23 @@ where
         }
     }
 
+    #[inline]
+    pub fn decode_bit_check(&mut self, prob: u16) -> io::Result<bool> {
```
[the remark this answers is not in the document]
cccs-sadugas: The point of this function is to avoid updating prob (i.e. it's not declared as a `&mut prob`). Otherwise, it does not much differently than the non-`_check` version.

Why not update prob while checking?

Updating probs could cause a scenario where the state is non-recoverable. I think the updated prob value is only meant to be used in the next iteration of the loop anyway, so it should be fine for checking this iteration.

See reference implementation [LzmaDec_TryDummy](https://github.com/OISF/libhtp/blob/0.5.x/htp/lzma/LzmaDec.c#L690).

src/decode/stream.rs
cccs-sadugas: I had to implement `Display` for `lzma_rs::error::Error`. I put this in a separate PR #53 .

src/decode/lzbuffer.rs
```diff
@@ -1,38 +1,51 @@
...
 }
 
 // An accumulating buffer for LZ sequences
-pub struct LZAccumBuffer<'a, W>
+pub struct LZAccumBuffer<W>
 where
-    W: 'a + io::Write,
+    W: io::Write,
 {
-    stream: &'a mut W, // Output sink
```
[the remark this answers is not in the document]
cccs-sadugas: Good idea. See PR #54.

src/decode/stream.rs
cccs-sadugas: Unfortunately that would require adding a type parameter to `LZMAError`, i.e. `LZMAError<W>`. I think a less intrusive way would be to create an error type inside of `stream.rs` and returning that:
```
Result<State<W>, StreamStateError<W>>
```

src/decode/stream.rs
cccs-sadugas: Added a test file and include it with `include_bytes!("../../tests/files/small.txt")` given that it's a small input.

benches/lzma.rs
cccs-sadugas: Good catch.

PR #53, opened 2020-07-14
Implement Display for Error

cccs-sadugas:
### Pull Request Overview

Useful for printing out error messages to users.

This is needed by future versions of #51 .

### Testing Strategy

Already covered by existing test cases

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,6 @@
 //! Error handling.
 
+use std::fmt::Display;
 use std::io;
 use std::result;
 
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -22,3 +23,37 @@ impl From<io::Error> for Error {
         Error::IOError(e)
     }
 }
+
+impl Display for Error {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Error::IOError(e) => e.fmt(fmt),
+            Error::LZMAError(e) | Error::XZError(e) => e.fmt(fmt),
```
gendx, line 31: The original error type (io, lzma or xz) should be formatted as well.

cccs-sadugas: > 
> 
> I think it would also be useful to implement the [`std::error::Error` trait](https://doc.rust-lang.org/std/error/trait.Error.html) (essentially implementing the `source()` method to return `Some(e)` in case of IO error and `None` otherwise).

I was thinking the same. See the latest commit.

PR #54, opened 2020-07-14
lzbuffer: own output sink instead of holding ref

cccs-sadugas:
### Pull Request Overview

Changes `LZBuffer` trait to consume the output sink instead of holding a
reference to it. This makes it easier to store the sink and avoids
self-referential structs. It also makes sense for the buffer to own the
sink while it is performing decompression.

This also adds the methods `get_output` and `get_output_mut` to access
the output sink.

This pull request is needed by future versions of PR #51.

### Testing Strategy

Should already be covered by existing tests.

Benchmarks:

**master**
```
test compress_65536                  ... bench:   3,187,997 ns/iter (+/- 863,604)
test compress_empty                  ... bench:       1,873 ns/iter (+/- 615)
test compress_hello                  ... bench:       2,072 ns/iter (+/- 608)
test decompress_after_compress_65536 ... bench:   3,624,633 ns/iter (+/- 1,528,450)
test decompress_after_compress_empty ... bench:       4,493 ns/iter (+/- 1,926)
test decompress_after_compress_hello ... bench:       5,452 ns/iter (+/- 5,584)
test decompress_big_file             ... bench:   8,809,873 ns/iter (+/- 5,368,703)
test decompress_huge_dict            ... bench:       4,430 ns/iter (+/- 2,042)
```

**owned_output_sink**
```
test compress_65536                  ... bench:   2,928,550 ns/iter (+/- 1,354,373)
test compress_empty                  ... bench:       1,513 ns/iter (+/- 512)
test compress_hello                  ... bench:       2,410 ns/iter (+/- 742)
test decompress_after_compress_65536 ... bench:   3,295,758 ns/iter (+/- 1,131,631)
test decompress_after_compress_empty ... bench:       3,893 ns/iter (+/- 1,449)
test decompress_after_compress_hello ... bench:       4,687 ns/iter (+/- 1,630)
test decompress_big_file             ... bench:   7,352,772 ns/iter (+/- 2,684,627)
test decompress_huge_dict            ... bench:       5,731 ns/iter (+/- 3,119)
```

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -3,6 +3,7 @@ use crate::decode::rangecoder;
 use crate::error;
 use byteorder::{LittleEndian, ReadBytesExt};
 use std::io;
+use std::marker::PhantomData;
 
 use crate::decompress::Options;
 use crate::decompress::UnpackedSize;
@@ -97,10 +98,12 @@ impl LZMAParams {
     }
 }
 
-pub struct DecoderState<LZB>
+pub struct DecoderState<W, LZB>
 where
-    LZB: lzbuffer::LZBuffer,
+    W: io::Write,
+    LZB: lzbuffer::LZBuffer<W>,
 {
+    _phantom: PhantomData<W>,
     pub output: LZB,
     // most lc significant bits of previous byte are part of the literal context
     pub lc: u32, // 0..8
@@ -125,17 +128,18 @@ where
 }
 
 // Initialize decoder with accumulating buffer
-pub fn new_accum<'a, W>(
-    output: lzbuffer::LZAccumBuffer<'a, W>,
+pub fn new_accum<W>(
+    output: lzbuffer::LZAccumBuffer<W>,
     lc: u32,
     lp: u32,
     pb: u32,
     unpacked_size: Option<u64>,
-) -> DecoderState<lzbuffer::LZAccumBuffer<'a, W>>
+) -> DecoderState<W, lzbuffer::LZAccumBuffer<W>>
 where
     W: io::Write,
 {
     DecoderState {
+        _phantom: PhantomData,
         output,
         lc,
         lp,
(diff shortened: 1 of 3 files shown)


src/decode/lzma.rs
```diff
@@ -97,10 +98,12 @@ impl LZMAParams {
...
 }
 
-pub struct DecoderState<LZB>
+pub struct DecoderState<W, LZB>
 where
-    LZB: lzbuffer::LZBuffer,
+    W: io::Write,
+    LZB: lzbuffer::LZBuffer<W>,
 {
+    _phantom: PhantomData<W>,
```
gendx, line 106: I'm not sure I understand why this `PhantomData` is needed. Could you briefly explain?

cccs-sadugas: It's because the `DecoderState` is parameterized over both `W` and `LZB` but the type `W` has to be used so the compiler doesn't complain.

If we declare the struct like so, we get an error:
```rust
pub struct DecoderState<LZB>
where
    LZB: lzbuffer::LZBuffer<W>
```
```
error[E0412]: cannot find type `W` in this scope
   --> src/decode/lzma.rs:103:29
    |
100 | pub struct DecoderState<LZB>
    |                            - help: you might be missing a type parameter: `, W`
...
103 |     LZB: lzbuffer::LZBuffer<W>,
    |                             ^ not found in this scope
```

If we add `W` we get an unused error:
```rust
pub struct DecoderState<W, LZB>
where
    W: io::Write,
    LZB: lzbuffer::LZBuffer<W>
```
```
error[E0392]: parameter `W` is never used
   --> src/decode/lzma.rs:100:25
    |
100 | pub struct DecoderState<W, LZB>
    |                         ^ unused parameter
    |
    = help: consider removing `W`, referring to it in a field, or using a marker such as `std::marker::PhantomData`
```

PR #55, opened 2020-07-14
Streaming Decompressor v2

cccs-sadugas:
### Pull Request Overview

Changes since last PR:
- Rebased from master to include memlimit changes.
- Published PRs for decoupled features.
- Addressed Travis CI build errors.
- Addressed code review comments.

### Testing Strategy

Link to previous PR: #51 

### TODO
- [x] Merge dependant PR #53.
- [x] Merge dependant PR #54.

### DIFF
--- a/src/decode/stream.rs
+++ b/src/decode/stream.rs
@@ -1,0 +1,497 @@
+use crate::decode::lzbuffer::{LZBuffer, LZCircularBuffer};
+use crate::decode::lzma::{
+    new_circular, new_circular_with_memlimit, DecoderState, LZMAParams, MAX_HEADER_LEN, START_BYTES,
+};
+use crate::decode::rangecoder::RangeDecoder;
+use crate::decompress::Options;
+use crate::error::Error;
+use std::io::{BufRead, Cursor, Read, Write};
+
+/// Maximum number of bytes to buffer while reading the header.
+const MAX_TMP_LEN: usize = MAX_HEADER_LEN + START_BYTES;
+
+/// Internal state of this streaming decoder. This is needed because we have to
+/// initialize the stream before processing any data.
+#[derive(Debug)]
+enum State<W>
+where
+    W: Write,
+{
+    /// Stream is initialized but header values have not yet been read.
+    Init(W),
+    /// Header values have been read and the stream is ready to process more data.
+    Run(RunState<W>),
+}
+
+/// Error type used to return the currently owned state along with the error on failure.
+struct StreamStateError<W>
+where
+    W: Write,
+{
+    /// State that was owned while the error occured
+    state: Option<State<W>>,
+    /// IO error that occured in that state
+    error: Error,
+}
+
+impl<W> StreamStateError<W>
+where
+    W: Write,
+{
+    /// Creates a new StreamStateError
+    fn new(state: Option<State<W>>, error: Error) -> Self {
+        Self { state, error }
+    }
+}
+
+/// Structures needed while decoding data.
+struct RunState<W>
+where
+    W: Write,
+{
+    decoder: DecoderState<W, LZCircularBuffer<W>>,
+    range: u32,
+    code: u32,
+}
+
+impl<W> RunState<W>
+where
+    W: Write,
+{
+    fn new(decoder: DecoderState<W, LZCircularBuffer<W>>, range: u32, code: u32) -> Self {
+        Self {
+            decoder,
+            range,
+            code,
+        }
+    }
+}
+
+impl<W> std::fmt::Debug for RunState<W>
+where
+    W: Write,
+{
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        fmt.debug_struct("RunState")
+            .field("range", &self.range)
+            .field("code", &self.code)
+            .finish()
+    }
+}
+
+/// Lzma decompressor that can process multiple chunks of data using the
+/// `std::io::Write` interface.
+pub struct Stream<W>
+where
+    W: Write,
+{
+    /// Temporary buffer to hold data while the header is being read.
+    tmp: [u8; MAX_TMP_LEN],
+    /// How many bytes of the temp buffer are in use.
+    tmp_len: usize,
+    /// Whether the stream is initialized and ready to process data.
+    /// An `Option` is used to avoid interior mutability when updating the state.
+    state: Option<State<W>>,
+    /// Options given when a stream is created.
+    options: Options,
+}
+
+impl<W> Stream<W>
+where
+    W: Write,
+{
+    /// Initialize the stream. This will consume the `output` which is the sink
+    /// implementing `std::io::Write` that will receive decompressed bytes.
+    pub fn new(output: W) -> Self {
+        Self::new_with_options(Options::default(), output)
+    }
+
+    /// Initialize the stream with the given `options`. This will consume the
+    /// `output` which is the sink implementing `std::io::Write` that will
+    /// receive decompressed bytes.
+    pub fn new_with_options(options: Options, output: W) -> Self {
+        Self {
+            tmp: [0; MAX_TMP_LEN],
+            tmp_len: 0,
+            state: Some(State::Init(output)),
+            options,
+        }
+    }
+
+    /// Get a reference to the output sink
+    pub fn get_ref(&self) -> Option<&W> {
+        match &self.state {
+            Some(State::Init(output)) => Some(&output),
+            Some(State::Run(state)) => Some(state.decoder.output.get_output()),
+            None => None,
+        }
+    }
+
+    /// Get a mutable reference to the output sink
+    pub fn get_mut(&mut self) -> Option<&mut W> {
+        match &mut self.state {
+            Some(State::Init(output)) => Some(output),
+            Some(State::Run(state)) => Some(state.decoder.output.get_output_mut()),
+            None => None,
+        }
+    }
+
+    /// Consumes the stream and returns the output sink. This also makes sure
+    /// we have properly reached the end of the stream.
+    pub fn finish(mut self) -> crate::error::Result<W> {
+        if let Some(state) = self.state.take() {
+            match state {
+                State::Init(output) => {
+                    if self.tmp_len > 0 {
+                        Err(Error::LZMAError("failed to read header".to_string()))
+                    } else {
+                        Ok(output)
+                    }
+                }
+                State::Run(mut state) => {
+                    if !self.options.allow_incomplete {
+                        // Process one last time with empty input to force end of
+                        // stream checks
+                        let mut stream = Cursor::new(&self.tmp[0..self.tmp_len]);
+                        let mut range_decoder =
+                            RangeDecoder::from_parts(&mut stream, state.range, state.code);
+                        state.decoder.process(&mut range_decoder)?;
+                    }
+                    let output = state.decoder.output.finish()?;
+                    Ok(output)
+                }
+            }
+        } else {
+            // this will occur if a call to `write()` fails
+            Err(Error::LZMAError(
+                "can't finish stream because of previous write error".to_string(),
+            ))
+        }
+    }
+
+    /// Attempts to read the header and transition into a running state.
+    ///
+    /// This function will consume the state, returning the next state on both
+    /// error and success.
+    fn read_header<R: BufRead>(
+        output: W,
+        mut input: &mut R,
+        options: &Options,
+    ) -> Result<State<W>, StreamStateError<W>> {
+        let params = match LZMAParams::read_header(&mut input, options) {
+            Ok(params) => params,
+            Err(e) => {
+                return Err(StreamStateError::new(Some(State::Init(output)), e));
+            }
+        };
+
+        let len = match input.fill_buf() {
+            Ok(val) => val,
+            Err(_) => {
+                return Err(StreamStateError::new(
+                    Some(State::Init(output)),
+                    Error::LZMAError("need more input".to_string()),
+                ));
+            }
+        }
+        .len();
+
+        if len < START_BYTES {
+            return Err(StreamStateError::new(
+                Some(State::Init(output)),
+                Error::LZMAError("need more input".to_string()),
+            ));
+        };
+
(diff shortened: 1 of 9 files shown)


cccs-sadugas:
Should I squash all the commits into a single one?

gendx:
> Should I squash all the commits into a single one?

It would be useful once the other two pull requests are merged.

gendx:
I have now merged the two related pull requests.

cccs-sadugas:
I have rebased and squashed. The new benchmarks are:

**master**
```
test compress_65536                  ... bench:   2,489,961 ns/iter (+/- 195,741)
test compress_empty                  ... bench:       1,464 ns/iter (+/- 267)
test compress_hello                  ... bench:       2,079 ns/iter (+/- 292)
test decompress_after_compress_65536 ... bench:   2,768,485 ns/iter (+/- 616,316)
test decompress_after_compress_empty ... bench:       3,346 ns/iter (+/- 674)
test decompress_after_compress_hello ... bench:       3,851 ns/iter (+/- 590)
test decompress_big_file             ... bench:   6,503,570 ns/iter (+/- 1,350,205)
test decompress_huge_dict            ... bench:       3,959 ns/iter (+/- 758)
```

**streaming-decompressor-v2**
```
test compress_65536                  ... bench:   2,611,232 ns/iter (+/- 467,790)
test compress_empty                  ... bench:       1,459 ns/iter (+/- 198)
test compress_hello                  ... bench:       2,128 ns/iter (+/- 401)
test decompress_after_compress_65536 ... bench:   2,569,576 ns/iter (+/- 272,391)
test decompress_after_compress_empty ... bench:       3,351 ns/iter (+/- 372)
test decompress_after_compress_hello ... bench:       4,050 ns/iter (+/- 463)
test decompress_big_file             ... bench:   6,325,453 ns/iter (+/- 727,326)
test decompress_huge_dict            ... bench:       4,162 ns/iter (+/- 760)
test decompress_stream_big_file      ... bench:   6,160,083 ns/iter (+/- 726,963)
```

cccs-sadugas:
> Can you add unit tests that the decompressed output is the same via streaming vs. with the direct interface?

Is there anything you were looking for in particular? All tests in `tests/stream.rs` assert the expected originally decompressed data is equal to the output of the stream decompressor.

tests/stream.rs
```diff
@@ -0,0 +1,274 @@
+#[cfg(feature = "enable_logging")]
```
gendx, line 1: These tests are good. It would be nice to somehow factor in the code with `tests/lzma.rs` to avoid code duplication, and make sure that new tests are automatically picked up on both "normal" and "stream" modes.

I'm not sure how to best do that yet (having tests generic over some trait for example), so it shouldn't block this pull request, but any idea is appreciated of course.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+/// Error type used to return the currently owned state along with the error on failure.
+struct StreamStateError<W>
+where
+    W: Write,
+{
+    /// State that was owned while the error occured
+    state: Option<State<W>>,
+    /// IO error that occured in that state
```
gendx, line 33: Is it necessarily an IO error?

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+
+impl<W> StreamStateError<W>
+where
+    W: Write,
+{
+    /// Creates a new StreamStateError
+    fn new(state: Option<State<W>>, error: Error) -> Self {
+        Self { state, error }
+    }
+}
```
gendx, line 45: Given that this is only used internally, I don't think this `new` function makes things clearer. In general `StreamStateError::new(foo, bar)` is less readable than `StreamStateError { state: foo, error: bar }`, because the function doesn't retain the field names.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+    W: Write,
+{
+    fn new(decoder: DecoderState<W, LZCircularBuffer<W>>, range: u32, code: u32) -> Self {
+        Self {
+            decoder,
+            range,
+            code,
+        }
+    }
+}
```
gendx, line 68: Same here.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+
+impl<W> std::fmt::Debug for RunState<W>
```
gendx, line 70: I'd import `use std::fmt::Debug` at the file level.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// Initialize the stream. This will consume the `output` which is the sink
+    /// implementing `std::io::Write` that will receive decompressed bytes.
+    pub fn new(output: W) -> Self {
+        Self::new_with_options(Options::default(), output)
+    }
+
+    /// Initialize the stream with the given `options`. This will consume the
+    /// `output` which is the sink implementing `std::io::Write` that will
+    /// receive decompressed bytes.
+    pub fn new_with_options(options: Options, output: W) -> Self {
```
gendx, line 112: The non-stream API takes an `&Options` as input. We should be consistent.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+        Self {
+            tmp: [0; MAX_TMP_LEN],
+            tmp_len: 0,
+            state: Some(State::Init(output)),
+            options,
+        }
+    }
+
+    /// Get a reference to the output sink
+    pub fn get_ref(&self) -> Option<&W> {
```
gendx, line 122: The name should be clearer. From the call side it's unclear a reference to what is obtained.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+        match &self.state {
+            Some(State::Init(output)) => Some(&output),
```
gendx, line 124: `self.state.map(|s| match s { ... })` would be more idiomatic.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+    fn read_header<R: BufRead>(
+        output: W,
+        mut input: &mut R,
+        options: &Options,
+    ) -> Result<State<W>, StreamStateError<W>> {
+        let params = match LZMAParams::read_header(&mut input, options) {
+            Ok(params) => params,
+            Err(e) => {
+                return Err(StreamStateError::new(Some(State::Init(output)), e));
+            }
```
gendx, line 185: This pattern would be more idiomatic as `let params = read_header(...).map_err(|e| ... )?;`

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+        };
+
+        let len = match input.fill_buf() {
+            Ok(val) => val,
+            Err(_) => {
```
gendx, line 190: Here as well `map_err` would be clearer.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+                return Err(StreamStateError::new(
+                    Some(State::Init(output)),
+                    Error::LZMAError("need more input".to_string()),
+                ));
+            }
+        }
+        .len();
+
+        if len < START_BYTES {
```
gendx, line 199: Is it an error? As far as I'm aware, `fill_buf` can return as few as 1 byte at a time, but it doesn't mean that `START_BYTES` aren't available afterwards.

From the documentation (https://doc.rust-lang.org/std/io/trait.BufRead.html#tymethod.fill_buf).

> An empty buffer returned indicates that the stream has reached EOF.

It would actually be nice to have some unit tests for this streaming API with some `BufRead` implementation that returns 1 byte at a time. Although that would be inefficient, that would also test the limits of the implementation.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+
+/// Internal state of this streaming decoder. This is needed because we have to
+/// initialize the stream before processing any data.
+#[derive(Debug)]
+enum State<W>
+where
+    W: Write,
+{
+    /// Stream is initialized but header values have not yet been read.
+    Init(W),
```
gendx, line 21: I'd name these states `Header` and `Data` for clarity.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+            code,
+        }
+    }
+}
+
+impl<W> std::fmt::Debug for RunState<W>
+where
+    W: Write,
+{
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
```
gendx, line 74: Do we need Debug formatting in the general case, or should it be restricted to `cfg(feature = "enable_logging")` and/or `cfg(test)`?

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+
+    /// Attempts to read the header and transition into a running state.
+    ///
+    /// This function will consume the state, returning the next state on both
+    /// error and success.
+    fn read_header<R: BufRead>(
+        output: W,
+        mut input: &mut R,
+        options: &Options,
+    ) -> Result<State<W>, StreamStateError<W>> {
```
gendx, line 180: To me, this `Result<State<W>, StreamStateError<W>>` doesn't describe well what's returned by this function. To me, there are 3 options:
- Success with a state,
- Need more data (isn't it a success with some state as well?),
- A real error (`crate::error::Error`).

I don't think conflating "need more data" into the `StreamStateError` is suitable. I think the `Ok` side should represent the "need more data" status - then the caller can handle that and just forward a "real error" if it happens.

src/decode/stream.rs
```diff
@@ -0,0 +1,497 @@
...
+/// Lzma decompressor that can process multiple chunks of data using the
+/// `std::io::Write` interface.
+pub struct Stream<W>
+where
+    W: Write,
+{
+    /// Temporary buffer to hold data while the header is being read.
+    tmp: [u8; MAX_TMP_LEN],
+    /// How many bytes of the temp buffer are in use.
+    tmp_len: usize,
```
gendx, line 91: I think this pair of fields would deserve its own helper struct, to make sure that things are properly encapsulated, and that we can't read beyond the current `tmp_len`.

tests/stream.rs
cccs-sadugas: Good idea. I made the common test functions like `round_trip` test both the non-streaming and streaming implementations and got rid of the stream.rs test module.

src/decode/stream.rs
cccs-sadugas: Looks like it is now the crate's error type.

src/decode/stream.rs
cccs-sadugas: Generally I like adding `new` functions to initialize all structs. This makes it easy to change initialization in the future if the struct changes.

However, I also don't mind switching it to struct initialization.

src/decode/stream.rs
cccs-sadugas: The build fails without debug implementations:

```
error: type does not implement `std::fmt::Debug`; consider adding `#[derive(Debug)]` or a manual implementation
  --> src/decode/stream.rs:48:1
   |
48 | / pub struct Stream<W>
49 | | where
50 | |     W: Write,
51 | | {
...  |
60 | |     options: Options,
61 | | }
   | |_^
   |
note: the lint level is defined here
  --> src/lib.rs:4:9
   |
4  | #![deny(missing_debug_implementations)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: build failed
```

src/decode/stream.rs
cccs-sadugas: Fair point. I was able to return a `crate::error::Result<State<W>>` and use state transitions to identify the case when we need more data:
- `Ok(State::Header(_))` means we need more data.
- `Ok(State::Data(_))` means we successfully transitioned states.
- `Err(_)` is a real error.

src/decode/stream.rs
cccs-sadugas: I was debating this myself. Thinking of using something like std::io::Cursor to keep track of the `tmp_len` "position" in the buffer.

src/decode/stream.rs
cccs-sadugas: I'll change this for consistency. Unfortunately it results in an extra clone when calling `new()`:
```rust
pub fn new(output: W) -> Self {
    Self::new_with_options(&Options::default(), output)
}

pub fn new_with_options(options: &Options, output: W) -> Self {
    Self {
        tmp: [0; MAX_TMP_LEN],
        tmp_len: 0,
        state: Some(State::Header(output)),
        options: options.clone(),
    }
}
```

src/decode/stream.rs
cccs-sadugas: I changed it to `get_output()` and `get_output_mut()` to be consistent with `LZBuffer`.

src/decode/stream.rs
cccs-sadugas: This doesn't apply now that we are no longer returning a `StreamStateError`. I did change it to an `if let Ok(...)`.

src/decode/stream.rs
cccs-sadugas: Same as above, using `if let Ok(...)` now.

src/decode/stream.rs
cccs-sadugas: There were a couple of issues with this function that I revisited. Mainly, we need to distinguish between a fatal error and "need more bytes". That's why I was doing a length check before continuing. However, this length check should have been done before calling `read_header()`. Doing so was not possible because the compiler reasonably complains that `output` is being moved multiple times.

Instead of doing a length check, I added an `Error::HeaderTooShort` (one step towards #6 maybe) so that we can distinguish between errors and handle the "need more bytes" case.

I also added `into_output()` to get back the output sink that was moved into the `DecoderState`'s `LZBuffer`.

Lastly, I added a test case `test_bad_header` to show that we properly fail on an invalid header.

cccs-sadugas: Regarding the unit tests, I do have a couple of tests in `stream.rs` to test resilience like this. `test_stream_chunked` will test every single chunk size from 1 on the whole input size.

cccs-sadugas:
Thanks for the thorough comments @gendx . I pushed changes to #58.

PR #58, opened 2020-08-10
Streaming Decompressor v3

cccs-sadugas:
### Pull Request Overview

Changes since last PR:
- Added feature gate for `"stream"`.
- Removed code duplication caused by `_check()` functions, using an `update` flag instead.
- Merged stream test module with lzma to remove duplication.
- Changed `tmp`, `tmp_len` buffers to use a `Cursor`.
- Addressed other code review comments.

### Testing Strategy

Link to previous PR: #55

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -8,6 +8,24 @@ use std::marker::PhantomData;
 use crate::decompress::Options;
 use crate::decompress::UnpackedSize;
 
+/// Maximum input data that can be processed in one iteration
+const MAX_REQUIRED_INPUT: usize = 20;
+
+/// Processing mode for decompression.
+///
+/// Tells the decompressor if we should expect more data after parsing the
+/// current input.
+#[derive(Debug, PartialEq)]
+pub enum Mode {
+    /// Streaming mode. Process the input bytes but assume there will be more
+    /// chunks of input data to receive in future calls to `process_mode()`.
+    Run,
+    /// Sync mode. Process the input bytes and confirm end of stream has been reached.
+    /// Use this mode if you are processing a fixed buffer of compressed data, or after
+    /// using `Mode::Run` to check for the end of stream.
+    Finish,
+}
+
 pub struct LZMAParams {
     // most lc significant bits of previous byte are part of the literal context
     lc: u32, // 0..8
@@ -24,9 +42,7 @@ impl LZMAParams {
         R: io::BufRead,
     {
         // Properties
-        let props = input
-            .read_u8()
-            .map_err(|e| error::Error::LZMAError(format!("LZMA header too short: {}", e)))?;
+        let props = input.read_u8().map_err(error::Error::HeaderTooShort)?;
 
         let mut pb = props as u32;
         if pb >= 225 {
@@ -46,7 +62,7 @@ impl LZMAParams {
         // Dictionary
         let dict_size_provided = input
             .read_u32::<LittleEndian>()
-            .map_err(|e| error::Error::LZMAError(format!("LZMA header too short: {}", e)))?;
+            .map_err(error::Error::HeaderTooShort)?;
         let dict_size = if dict_size_provided < 0x1000 {
             0x1000
         } else {
@@ -58,9 +74,9 @@ impl LZMAParams {
         // Unpacked size
         let unpacked_size: Option<u64> = match options.unpacked_size {
             UnpackedSize::ReadFromHeader => {
-                let unpacked_size_provided = input.read_u64::<LittleEndian>().map_err(|e| {
-                    error::Error::LZMAError(format!("LZMA header too short: {}", e))
-                })?;
+                let unpacked_size_provided = input
+                    .read_u64::<LittleEndian>()
+                    .map_err(error::Error::HeaderTooShort)?;
                 let marker_mandatory: bool = unpacked_size_provided == 0xFFFF_FFFF_FFFF_FFFF;
                 if marker_mandatory {
                     None
@@ -69,7 +85,9 @@ impl LZMAParams {
                 }
             }
             UnpackedSize::ReadHeaderButUseProvided(x) => {
-                input.read_u64::<LittleEndian>()?;
+                input
+                    .read_u64::<LittleEndian>()
+                    .map_err(error::Error::HeaderTooShort)?;
                 x
             }
             UnpackedSize::UseProvided(x) => x,
@@ -95,6 +113,9 @@ where
     LZB: lzbuffer::LZBuffer<W>,
 {
     _phantom: PhantomData<W>,
+    // buffer input data here if we need more for decompression, 20 is the max
+    // number of bytes that can be consumed during one iteration
+    tmp: std::io::Cursor<[u8; MAX_REQUIRED_INPUT]>,
     pub output: LZB,
     // most lc significant bits of previous byte are part of the literal context
     pub lc: u32, // 0..8
@@ -131,6 +152,7 @@ where
 {
     DecoderState {
         _phantom: PhantomData,
+        tmp: std::io::Cursor::new([0; MAX_REQUIRED_INPUT]),
         output,
         lc,
         lp,
@@ -181,6 +203,7 @@ where
             params.dict_size as usize,
             memlimit,
         ),
+        tmp: std::io::Cursor::new([0; MAX_REQUIRED_INPUT]),
         lc: params.lc,
         lp: params.lp,
         pb: params.pb,
@@ -237,23 +260,38 @@ where
         &mut self,
         rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
     ) -> error::Result<()> {
-        loop {
-            if let Some(unpacked_size) = self.unpacked_size {
-                if self.output.len() as u64 >= unpacked_size {
-                    break;
-                }
-            } else if rangecoder.is_finished_ok()? {
-                break;
-            }
+        self.process_mode(rangecoder, Mode::Finish)
+    }
 
-            let pos_state = self.output.len() & ((1 << self.pb) - 1);
+    #[cfg(feature = "stream")]
+    pub fn process_stream<'a, R: io::BufRead>(
+        &mut self,
+        rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
+    ) -> error::Result<()> {
+        self.process_mode(rangecoder, Mode::Run)
+    }
 
-            // Literal
-            if !rangecoder.decode_bit(
-                // TODO: assumes pb = 2 ??
-                &mut self.is_match[(self.state << 4) + pos_state],
-            )? {
-                let byte: u8 = self.decode_literal(rangecoder)?;
+    /// Process the next iteration of the loop.
+    ///
+    /// If the update flag is true, the decoder's state will be updated.
+    ///
+    /// Returns true if we should continue processing the loop, false otherwise.
+    fn process_next_inner<'a, R: io::BufRead>(
+        &mut self,
+        rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
+        update: bool,
+    ) -> error::Result<bool> {
+        let pos_state = self.output.len() & ((1 << self.pb) - 1);
+
+        // Literal
+        if !rangecoder.decode_bit(
+            // TODO: assumes pb = 2 ??
+            &mut self.is_match[(self.state << 4) + pos_state],
+            update,
+        )? {
+            let byte: u8 = self.decode_literal(rangecoder, update)?;
+
+            if update {
                 lzma_debug!("Literal: {}", byte);
                 self.output.append_literal(byte)?;
 
@@ -264,35 +302,40 @@ where
                 } else {
                     self.state - 6
                 };
-                continue;
             }
+            return Ok(true);
+        }
 
-            // LZ
-            let mut len: usize;
-            // Distance is repeated from LRU
-            if rangecoder.decode_bit(&mut self.is_rep[self.state])? {
-                // dist = rep[0]
-                if !rangecoder.decode_bit(&mut self.is_rep_g0[self.state])? {
-                    // len = 1
-                    if !rangecoder
-                        .decode_bit(&mut self.is_rep_0long[(self.state << 4) + pos_state])?
-                    {
-                        // update state (short rep)
+        // LZ
+        let mut len: usize;
+        // Distance is repeated from LRU
+        if rangecoder.decode_bit(&mut self.is_rep[self.state], update)? {
+            // dist = rep[0]
+            if !rangecoder.decode_bit(&mut self.is_rep_g0[self.state], update)? {
+                // len = 1
+                if !rangecoder.decode_bit(
+                    &mut self.is_rep_0long[(self.state << 4) + pos_state],
+                    update,
+                )? {
+                    // update state (short rep)
+                    if update {
                         self.state = if self.state < 7 { 9 } else { 11 };
                         let dist = self.rep[0] + 1;
                         self.output.append_lz(1, dist)?;
-                        continue;
                     }
-                // dist = rep[i]
+                    return Ok(true);
+                }
+            // dist = rep[i]
+            } else {
+                let idx: usize;
+                if !rangecoder.decode_bit(&mut self.is_rep_g1[self.state], update)? {
+                    idx = 1;
+                } else if !rangecoder.decode_bit(&mut self.is_rep_g2[self.state], update)? {
+                    idx = 2;
                 } else {
-                    let idx: usize;
-                    if !rangecoder.decode_bit(&mut self.is_rep_g1[self.state])? {
-                        idx = 1;
-                    } else if !rangecoder.decode_bit(&mut self.is_rep_g2[self.state])? {
-                        idx = 2;
-                    } else {
-                        idx = 3;
-                    }
+                    idx = 3;
+                }
+                if update {
                     // Update LRU
                     let dist = self.rep[idx];
                     for i in (0..idx).rev() {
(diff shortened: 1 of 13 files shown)


cccs-sadugas:
Benchmarks are:

**master**
```
test compress_65536                  ... bench:   3,136,090 ns/iter (+/- 243,071)
test compress_empty                  ... bench:       1,473 ns/iter (+/- 170)
test compress_hello                  ... bench:       2,149 ns/iter (+/- 215)
test decompress_after_compress_65536 ... bench:   3,846,980 ns/iter (+/- 554,928)
test decompress_after_compress_empty ... bench:       3,412 ns/iter (+/- 396)
test decompress_after_compress_hello ... bench:       4,375 ns/iter (+/- 514)
test decompress_big_file             ... bench:   7,479,885 ns/iter (+/- 855,966)
test decompress_huge_dict            ... bench:       4,333 ns/iter (+/- 685)
```
**streaming-decompressor-v3**
```
test compress_65536                  ... bench:   3,196,185 ns/iter (+/- 307,679)
test compress_empty                  ... bench:       1,418 ns/iter (+/- 161)
test compress_hello                  ... bench:       2,113 ns/iter (+/- 416)
test decompress_after_compress_65536 ... bench:   3,796,298 ns/iter (+/- 392,350)
test decompress_after_compress_empty ... bench:       3,325 ns/iter (+/- 370)
test decompress_after_compress_hello ... bench:       4,329 ns/iter (+/- 663)
test decompress_big_file             ... bench:   7,563,810 ns/iter (+/- 792,639)
test decompress_huge_dict            ... bench:       4,271 ns/iter (+/- 456)
test decompress_stream_big_file      ... bench:   7,194,940 ns/iter (+/- 647,558)
```

cccs-sadugas:
@gendx do you have an update on when you will be able to review this PR?

tests/lzma.rs
```diff
@@ -44,21 +60,30 @@ fn round_trip_with_options(
...
+
+    #[cfg(feature = "stream")]
+    // test streaming decompression
+    {
+        let mut stream = lzma_rs::decompress::Stream::new_with_options(decode_options, Vec::new());
+
+        if let Err(error) = stream.write_all(&compressed) {
+            // WriteZero could indicate that the unpacked_size was reached before the
+            // end of the stream
+            if std::io::ErrorKind::WriteZero != error.kind() {
```
gendx, line 80: Nit: `error.kind() != std::io::ErrorKind::WriteZero` is more readable.

tests/lzma.rs
```diff
@@ -44,21 +60,30 @@ fn round_trip_with_options(
...
+        lzma_rs::lzma_decompress_with_options(&mut bf, &mut decomp, decode_options).unwrap();
+        assert_eq!(decomp, x);
+    }
+
+    #[cfg(feature = "stream")]
+    // test streaming decompression
+    {
+        let mut stream = lzma_rs::decompress::Stream::new_with_options(decode_options, Vec::new());
+
+        if let Err(error) = stream.write_all(&compressed) {
```
gendx, line 77: Why is there an error check here but not in `round_trip_no_options`?

tests/lzma.rs
```diff
@@ -75,30 +100,79 @@ fn round_trip_file(filename: &str) {
...
+
+    // test non-streaming decompression
+    {
+        let input = {
+            let mut input = Vec::new();
+            std::fs::File::open(compfile)
+                .unwrap()
+                .read_to_end(&mut input)
+                .unwrap();
+            input
```
gendx, line 120: Since this pattern appears twice already in this function, it would make sense to create a utility function for it.

tests/lzma.rs
```diff
@@ -75,30 +100,79 @@ fn round_trip_file(filename: &str) {
...
+        let mut stream = lzma_rs::decompress::Stream::new(Vec::new());
+
+        // read file in chunks
+        let mut tmp = [0u8; 1024];
+        while {
+            let n = compfile.read(&mut tmp).unwrap();
+            stream.write_all(&tmp[0..n]).unwrap();
+
+            n > 0
+        } {}
```
gendx, line 142: This logic is not readable. Can you use `if` together with `break` for the exit condition?

tests/lzma.rs
```diff
@@ -142,36 +216,36 @@ fn big_file() {
...
     #[cfg(feature = "enable_logging")]
     let _ = env_logger::try_init();
-    let mut x: &[u8] = b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x83\xff\
-                         \xfb\xff\xff\xc0\x00\x00\x00";
-    let mut decomp: Vec<u8> = Vec::new();
-    lzma_rs::lzma_decompress(&mut x, &mut decomp).unwrap();
-    assert_eq!(decomp, b"")
+    assert_decomp_eq(
+        b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x83\xff\
+                         \xfb\xff\xff\xc0\x00\x00\x00",
```
gendx, line 221: Nit: indentation should better align here.

tests/lzma.rs
```diff
@@ -142,36 +216,36 @@ fn big_file() {
...
-    let mut x: &[u8] = b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x24\x19\
-                         \x49\x98\x6f\x10\x19\xc6\xd7\x31\xeb\x36\x50\xb2\x98\x48\xff\xfe\
-                         \xa5\xb0\x00";
-    let mut decomp: Vec<u8> = Vec::new();
-    lzma_rs::lzma_decompress(&mut x, &mut decomp).unwrap();
-    assert_eq!(decomp, b"Hello world\x0a")
+    assert_decomp_eq(
+        b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x24\x19\
+        \x49\x98\x6f\x10\x19\xc6\xd7\x31\xeb\x36\x50\xb2\x98\x48\xff\xfe\
+        \xa5\xb0\x00",
```
gendx, line 233: Nit: indentation should better align here.

tests/lzma.rs
```diff
@@ -142,36 +216,36 @@ fn big_file() {
...
 
 #[test]
 fn decompress_huge_dict() {
     // Hello world with a dictionary of size 0x7F7F7F7F
     #[cfg(feature = "enable_logging")]
     let _ = env_logger::try_init();
-    let mut x: &[u8] = b"\x5d\x7f\x7f\x7f\x7f\xff\xff\xff\xff\xff\xff\xff\xff\x00\x24\x19\
+    assert_decomp_eq(
+        b"\x5d\x7f\x7f\x7f\x7f\xff\xff\xff\xff\xff\xff\xff\xff\x00\x24\x19\
                          \x49\x98\x6f\x10\x19\xc6\xd7\x31\xeb\x36\x50\xb2\x98\x48\xff\xfe\
```
gendx, line 245: Nit: indentation should better align here.

src/decode/rangecoder.rs
```diff
@@ -28,9 +28,51 @@ where
...
+        self.code = code;
+    }
+
+    pub fn range(&self) -> u32 {
+        self.range
+    }
+
+    pub fn code(&self) -> u32 {
+        self.code
+    }
```
gendx, line 54: I don't think these functions make things clear, because one doesn't know which one is which in the `u32, u32` pair. Making `range`/`code` public would be simpler and more readable.

src/decode/rangecoder.rs
```diff
@@ -28,9 +28,51 @@ where
...
+
+    pub fn buf(&mut self) -> io::Result<&[u8]> {
+        self.stream.fill_buf()
+    }
```
gendx, line 58: I think making `stream` public would also be more readable. A function named `buf()` doesn't really make things more readable than the original `fill_buf`.

src/decode/rangecoder.rs
```diff
@@ -28,9 +28,51 @@ where
...
+
+    pub fn remaining(&mut self) -> io::Result<usize> {
+        Ok(self.buf()?.len())
```
gendx, line 61: Is that correct? As far as I remember, `fill_buf` can return fewer bytes than will be available in total, so this function doesn't return the number of bytes remaining, but only a lower bound, which isn't clear from the function's name.

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+    ) -> error::Result<bool> {
+        self.process_next_inner(rangecoder, true)
+    }
+
+    /// Try to process the next iteration of the loop.
+    ///
+    /// This will check to see if there is enough data to consume and advance the
+    /// decompressor. Needed in streaming mode to avoid corrupting the state while
+    /// processing incomplete chunks of data.
+    #[allow(clippy::if_same_then_else)]
```
gendx, line 407: Why is this clippy lint allowed?

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+
+                if !self.process_next(&mut rangecoder)? {
+                    break;
+                };
+            }
+        }
+
         if let Some(len) = self.unpacked_size {
-            if self.output.len() as u64 != len {
+            if Mode::Finish == mode && self.output.len() as u64 != len {
```
gendx, line 510: Nit: `mode == Mode::Finish` is more readable.

src/decode/lzma.rs
```diff
@@ -8,6 +8,24 @@ use std::marker::PhantomData;
...
+/// Processing mode for decompression.
+///
+/// Tells the decompressor if we should expect more data after parsing the
+/// current input.
+#[derive(Debug, PartialEq)]
+pub enum Mode {
+    /// Streaming mode. Process the input bytes but assume there will be more
+    /// chunks of input data to receive in future calls to `process_mode()`.
+    Run,
+    /// Sync mode. Process the input bytes and confirm end of stream has been reached.
```
gendx, line 23: "Sync" => "Synchronous"

src/decode/lzma.rs
```diff
@@ -8,6 +8,24 @@ use std::marker::PhantomData;
...
+
+/// Processing mode for decompression.
+///
+/// Tells the decompressor if we should expect more data after parsing the
+/// current input.
+#[derive(Debug, PartialEq)]
+pub enum Mode {
+    /// Streaming mode. Process the input bytes but assume there will be more
+    /// chunks of input data to receive in future calls to `process_mode()`.
+    Run,
```
gendx, line 22: "Partial" would be clearer. "Run" doesn't mean much without context.

src/decode/lzma.rs
```diff
@@ -95,6 +113,9 @@ where
     LZB: lzbuffer::LZBuffer<W>,
 {
     _phantom: PhantomData<W>,
+    // buffer input data here if we need more for decompression, 20 is the max
+    // number of bytes that can be consumed during one iteration
```
gendx, line 117: This comment mentions "20" but there is no 20 here. Here it should refer to `MAX_REQUIRED_INPUT`, and the explanation for 20 should be given in `MAX_REQUIRED_INPUT`'s definition.

src/decode/lzma.rs
```diff
@@ -95,6 +113,9 @@ where
...
+    tmp: std::io::Cursor<[u8; MAX_REQUIRED_INPUT]>,
```
gendx, line 118: "tmp" isn't really meaningful. What does this contain? Maybe `partial_input_buf` would be clearer?

src/decode/lzma.rs
```diff
@@ -237,23 +260,38 @@ where
...
-            if !rangecoder.decode_bit(
-                // TODO: assumes pb = 2 ??
-                &mut self.is_match[(self.state << 4) + pos_state],
-            )? {
-                let byte: u8 = self.decode_literal(rangecoder)?;
+    /// Process the next iteration of the loop.
+    ///
+    /// If the update flag is true, the decoder's state will be updated.
+    ///
+    /// Returns true if we should continue processing the loop, false otherwise.
```
gendx, line 278: I think an enumeration would be clearer than a boolean to express the result. Something like `enum { Continue, Finished }`?

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                    return Ok(());
+                }
+
+                // Run the decompressor on the tmp buffer
+                let mut tmp_reader = io::Cursor::new(&tmp[0..self.tmp.position() as usize]);
+                let mut tmp_rangecoder = rangecoder::RangeDecoder::from_parts(
+                    &mut tmp_reader,
+                    rangecoder.range(),
+                    rangecoder.code(),
+                );
```
gendx, line 466: So, rather than exposing all fields and add all these from_parts/to_parts, what is really needed is some kind of `duplicate`/`clone` functionality for the `RangeDecoder` right? There could be a function specifically for `RangeCoder<Cursor<_>>`. I think that would be clearer to have that explicitly in the `RangeDecoder` struct rather than building and destructuring range coders here.

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                // Update the actual rangecoder
+                let (range, code) = tmp_rangecoder.into_parts();
+                rangecoder.set(range, code);
+
+                // Update tmp buffer
+                let end = self.tmp.position();
+                let new_len = end - tmp_reader.position();
+                self.tmp.get_mut()[0..new_len as usize]
+                    .copy_from_slice(&tmp[tmp_reader.position() as usize..end as usize]);
+                self.tmp.set_position(new_len);
```
gendx, line 478: Likewise, here we just want to update a range coder from another one.

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                Mode::Finish => rangecoder.is_finished_ok()?,
+            } {
+                break;
+            }
+
+            if self.tmp.position() as usize > 0 {
+                // Fill as much of the tmp buffer as possible
+                let start = self.tmp.position() as usize;
+                let bytes_read = rangecoder.read_into(&mut self.tmp.get_mut()[start..])?;
+                let bytes_read = if bytes_read < std::u64::MAX as usize {
```
gendx, line 436: I think it would be more readable to write:

```rust
if bytes_read == std::u64::MAX {
    return Err(...)
}
```

Also, why is the comparison on `usize` when the result is then converted to `u64`? And can the `bytes_read` ever reach this limit condition?

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                } else {
+                    return Err(error::Error::LZMAError(
+                        "Failed to convert integer to u64.".to_string(),
+                    ));
+                };
+                self.tmp.set_position(self.tmp.position() + bytes_read);
+                let tmp = *self.tmp.get_ref();
+
+                // Check if we need more data to advance the decompressor
+                if Mode::Run == mode
```
gendx, line 447: Same comment about `if variable == Constant` being more readable than the reverse. There are a few more in the code.

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                let new_len = end - tmp_reader.position();
+                self.tmp.get_mut()[0..new_len as usize]
+                    .copy_from_slice(&tmp[tmp_reader.position() as usize..end as usize]);
+                self.tmp.set_position(new_len);
+
+                if !res {
+                    break;
+                };
+            } else {
+                if (Mode::Run == mode) && (rangecoder.remaining()? < MAX_REQUIRED_INPUT) {
```
gendx, line 484: Same comment about `if variable == Constant` being more readable than the reverse. There are a few more in the code.

gendx, line 484: The parentheses are not necessary.

src/decode/lzma.rs
```diff
@@ -264,76 +302,212 @@ where
...
+                        let bytes_read = if bytes_read < std::u64::MAX as usize {
+                            bytes_read as u64
+                        } else {
+                            return Err(error::Error::LZMAError(
+                                "Failed to convert integer to u64.".to_string(),
+                            ));
+                        };
+                        self.tmp.set_position(bytes_read);
+                        return Ok(());
+                    }
```
gendx, line 500: This code looks similar to the above. Can you separate it into a utility function to avoid code duplication?

src/decode/stream.rs
```diff
@@ -0,0 +1,510 @@
+use crate::decode::lzbuffer::{LZBuffer, LZCircularBuffer};
```
gendx, line 1: Note: I didn't look in detail at this file (yet). But it is restricted to the `stream` feature.

gendx (changes requested):
Hello,

Sorry again for the delay in reviewing, a lot of things came up which meant I didn't have a lot of time in the past couple months, and this change is non-trivial to review.

I think we're getting there, most comments are for minor changes.

tests/lzma.rs
cccs-sadugas: You are correct. A simple `.unwrap()` will do. This was because a test case would previously fail with `WriteZero` for some `Options` when `unpacked_size` was reached.

tests/lzma.rs
cccs-sadugas: The rust-analyzer vscode plugin displays inlay hints and this was offsetting the first line in my IDE. Thankfully this can be disabled :) .

src/decode/lzma.rs
cccs-sadugas: I was able to remove it without any issues.

src/decode/lzma.rs
cccs-sadugas: The value 20 was derived from the comment mentioned here:
https://github.com/OISF/libhtp/blob/0.5.x/htp/lzma/LzmaDec.h#L48

src/decode/lzma.rs
cccs-sadugas: Yes, you are cloning the range and code but using a different value for `stream`.

This is a good idea. One issue that I had creating a clone function relates to the type `<R: io::BufRead>`.

Say we have a function called like so:
```rust
let tmp_buf = std::io::Cursor::new("test".as_bytes());
let new_rangecoder = rangecoder.clone_with(tmp_buf);
```

There is a constraint that the new_rangecoder must have the same concrete type `R` as the source of the clone. I.e. `clone_with` only works if both the `rangecoder.stream` and the `tmp_buf` are `R: std::io::Cursor<&[u8]>`. I'll see if anything can be done about this.

src/decode/lzma.rs
gendx: What I meant is that there is no `20` in the code here, just `MAX_REQUIRED_INPUT`, so the comment should use `MAX_REQUIRED_INPUT`rather than `20`. The explanation for `20` should be given where the `MAX_REQUIRED_INPUT` constant is defined.

src/decode/lzma.rs
cccs-sadugas: I've changed this to use `u64::try_from` which seems more appropriate.

src/decode/rangecoder.rs
cccs-sadugas: I removed this function since it was misleading and only used in one place. We are checking the number of currently available bytes to determine whether `try_process_next()` should be called.

tests/lzma.rs
cccs-sadugas: I take that back. After running the tests with `--all-features` it shows that `unpacked_size_write_none_to_header_and_use_provided_on_read` can return a `WriteZero` io error:

```
thread 'unpacked_size_write_none_to_header_and_use_provided_on_read' panicked
  at 'called `Result::unwrap()` on an `Err` value:
  Custom { kind: WriteZero, error: "failed to write whole buffer" }', tests/lzma.rs:88:39
```

This is expected behavior. We are encoding `0xFFFF_FFFF_FFFF` in the header but using the provided unpacked size during decoding. Therefore decoding will stop before the 5 to 6 byte end-of-stream marker is read.

The check is here but not in `round_trip_no_options` because it only applies when certain options are provided like in `unpacked_size_write_none_to_header_and_use_provided_on_read`. The `round_trip_no_options` function only uses default options.

I added more documentation and a match statement to make this more clear.

src/decode/lzma.rs
cccs-sadugas: I resolved the type issue but it is still not possible due to holding multiple references to the rangecoder. Copying the range and code avoids these issues.

src/decode/lzma.rs
cccs-sadugas: Similar to my other answer. It's not possible due to holding multiple references to the rangecoder. Copying the range and code avoids these issues.

cccs-sadugas:
Thanks for the review :). I published a new PR #59 with the requested changes.

I had some issues with the suggested changes for the `RangeDecoder::{from_parts, set}` functions so those are still up for discussion.

PR #59, opened 2020-10-05
Streaming Decompressor v4

cccs-sadugas:
### Pull Request Overview

Changes since last PR:
* removed YODA-style if-statements to improve readability
* removed do-while style logic to make a more readable `loop`
* removed confusing `RangeDecoder::remaining` function
* removed unnecessary clippy lint
* removed range checking when casting to `u64`
* renamed `Mode` to `ProcessingMode` and added a `ProcessingStatus` instead of using a `bool`
* renamed `tmp` to `partial_input_buf` to make it more clear
* fixed indentation on multi-line byte strings
* made `RangeDecoder` members public instead of having getters
* added more documentation to why `unpacked_size_write_none_to_header_and_use_provided_on_read` causes a `WriteZero` error
* added a test utility function for reading a file into a `Vec<u8>`
* added comment as to why `MAX_REQUIRED_INPUT` is 20 bytes
* added a utility function to `read_partial_input_buf`

### Testing Strategy

 Link to previous PR: #58

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -8,6 +8,36 @@ use std::marker::PhantomData;
 use crate::decompress::Options;
 use crate::decompress::UnpackedSize;
 
+/// Maximum input data that can be processed in one iteration.
+/// Libhtp uses the following equation to define the maximum number of bits
+/// for the worst case scenario:
+///   log2((2^11 / 31) ^ 22) + 26 < 134 + 26 = 160
+const MAX_REQUIRED_INPUT: usize = 20;
+
+/// Processing mode for decompression.
+///
+/// Tells the decompressor if we should expect more data after parsing the
+/// current input.
+#[derive(Debug, PartialEq)]
+pub enum ProcessingMode {
+    /// Streaming mode. Process the input bytes but assume there will be more
+    /// chunks of input data to receive in future calls to `process_mode()`.
+    Partial,
+    /// Synchronous mode. Process the input bytes and confirm end of stream has been reached.
+    /// Use this mode if you are processing a fixed buffer of compressed data, or after
+    /// using `Mode::Partial` to check for the end of stream.
+    Finish,
+}
+
+/// Result of the next iteration of processing.
+///
+/// Indicates whether processing should continue or is finished.
+#[derive(Debug, PartialEq)]
+enum ProcessingStatus {
+    Continue,
+    Finished,
+}
+
 pub struct LZMAParams {
     // most lc significant bits of previous byte are part of the literal context
     lc: u32, // 0..8
@@ -24,9 +54,7 @@ impl LZMAParams {
         R: io::BufRead,
     {
         // Properties
-        let props = input
-            .read_u8()
-            .map_err(|e| error::Error::LZMAError(format!("LZMA header too short: {}", e)))?;
+        let props = input.read_u8().map_err(error::Error::HeaderTooShort)?;
 
         let mut pb = props as u32;
         if pb >= 225 {
@@ -46,7 +74,7 @@ impl LZMAParams {
         // Dictionary
         let dict_size_provided = input
             .read_u32::<LittleEndian>()
-            .map_err(|e| error::Error::LZMAError(format!("LZMA header too short: {}", e)))?;
+            .map_err(error::Error::HeaderTooShort)?;
         let dict_size = if dict_size_provided < 0x1000 {
             0x1000
         } else {
@@ -58,9 +86,9 @@ impl LZMAParams {
         // Unpacked size
         let unpacked_size: Option<u64> = match options.unpacked_size {
             UnpackedSize::ReadFromHeader => {
-                let unpacked_size_provided = input.read_u64::<LittleEndian>().map_err(|e| {
-                    error::Error::LZMAError(format!("LZMA header too short: {}", e))
-                })?;
+                let unpacked_size_provided = input
+                    .read_u64::<LittleEndian>()
+                    .map_err(error::Error::HeaderTooShort)?;
                 let marker_mandatory: bool = unpacked_size_provided == 0xFFFF_FFFF_FFFF_FFFF;
                 if marker_mandatory {
                     None
@@ -69,7 +97,9 @@ impl LZMAParams {
                 }
             }
             UnpackedSize::ReadHeaderButUseProvided(x) => {
-                input.read_u64::<LittleEndian>()?;
+                input
+                    .read_u64::<LittleEndian>()
+                    .map_err(error::Error::HeaderTooShort)?;
                 x
             }
             UnpackedSize::UseProvided(x) => x,
@@ -95,6 +125,9 @@ where
     LZB: lzbuffer::LZBuffer<W>,
 {
     _phantom: PhantomData<W>,
+    // Buffer input data here if we need more for decompression. Up to
+    // MAX_REQUIRED_INPUT bytes can be consumed during one iteration.
+    partial_input_buf: std::io::Cursor<[u8; MAX_REQUIRED_INPUT]>,
     pub output: LZB,
     // most lc significant bits of previous byte are part of the literal context
     pub lc: u32, // 0..8
@@ -131,6 +164,7 @@ where
 {
     DecoderState {
         _phantom: PhantomData,
+        partial_input_buf: std::io::Cursor::new([0; MAX_REQUIRED_INPUT]),
         output,
         lc,
         lp,
@@ -181,6 +215,7 @@ where
             params.dict_size as usize,
             memlimit,
         ),
+        partial_input_buf: std::io::Cursor::new([0; MAX_REQUIRED_INPUT]),
         lc: params.lc,
         lp: params.lp,
         pb: params.pb,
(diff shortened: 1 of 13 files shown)


src/decode/lzma.rs
```diff
@@ -264,76 +315,206 @@ where
...
+
+    /// Utility function to read data into the partial input buffer.
+    fn read_partial_input_buf<'a, R: io::BufRead>(
+        &mut self,
+        rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
+    ) -> error::Result<()> {
+        // Fill as much of the tmp buffer as possible
+        let start = self.partial_input_buf.position() as usize;
+        let bytes_read =
+            rangecoder.read_into(&mut self.partial_input_buf.get_mut()[start..])? as u64;
```
cccs-sadugas, line 435: CI was failing because Rust 1.32 doesn't have `std::convert::TryFrom`. I think the cast to u64 is safe here because of the invariant `bytes_read <= MAX_REQUIRED_INPUT`.

cccs-sadugas:
Benchmarks are:

```
# master
test compress_65536                  ... bench:   1,902,882 ns/iter (+/- 526,492)
test compress_empty                  ... bench:       1,339 ns/iter (+/- 166)
test compress_hello                  ... bench:       1,783 ns/iter (+/- 217)
test decompress_after_compress_65536 ... bench:   1,824,803 ns/iter (+/- 416,725)
test decompress_after_compress_empty ... bench:       3,415 ns/iter (+/- 892)
test decompress_after_compress_hello ... bench:       3,913 ns/iter (+/- 660)
test decompress_big_file             ... bench:   6,070,123 ns/iter (+/- 657,337)
test decompress_huge_dict            ... bench:       3,853 ns/iter (+/- 453)

# branch with default features
test compress_65536                  ... bench:   1,843,732 ns/iter (+/- 166,190)
test compress_empty                  ... bench:       1,335 ns/iter (+/- 156)
test compress_hello                  ... bench:       1,749 ns/iter (+/- 188)
test decompress_after_compress_65536 ... bench:   2,049,825 ns/iter (+/- 394,655)
test decompress_after_compress_empty ... bench:       3,799 ns/iter (+/- 1,432)
test decompress_after_compress_hello ... bench:       4,451 ns/iter (+/- 2,081)
test decompress_big_file             ... bench:   6,201,524 ns/iter (+/- 697,083)
test decompress_huge_dict            ... bench:       3,869 ns/iter (+/- 546)

# branch with all features
test compress_65536                  ... bench:   2,440,331 ns/iter (+/- 188,993)
test compress_empty                  ... bench:       1,538 ns/iter (+/- 420)
test compress_hello                  ... bench:       2,210 ns/iter (+/- 647)
test decompress_after_compress_65536 ... bench:   3,717,059 ns/iter (+/- 608,489)
test decompress_after_compress_empty ... bench:       3,613 ns/iter (+/- 640)
test decompress_after_compress_hello ... bench:       4,572 ns/iter (+/- 458)
test decompress_big_file             ... bench:   7,830,967 ns/iter (+/- 875,957)
test decompress_huge_dict            ... bench:       4,578 ns/iter (+/- 648)
test decompress_stream_big_file      ... bench:   7,470,316 ns/iter (+/- 1,367,116)

```

gendx:
I'll re-review the changes, and perform benchmarks locally to confirm the performance impact.

From your results, it seems that there could be a slight performance regression even in non-streaming mode, maybe due to the changes in the range coder (`update` boolean parameter), in which case it could be worth putting that behind the `stream` feature as well to avoid the regression. If the regression is confirmed but the culprit is not the range coder, [producing a flamegraph](https://gendignoux.com/blog/2019/11/09/profiling-rust-docker-perf.html#profiling-rust-code) might be useful as well to see where the regression may come from.

I hope to get back with a more complete reply in the next few days.

GreenReaper:
@gendx: Do you have time to revisit this? https://github.com/ruffle-rs/ruffle/issues/405#issuecomment-600240689 indicates a need for streaming support prior to inclusion. Ruffle is getting more attention due to next month's demise of the Flash browser plugin. There isn't *that* much lzma-encoded Flash - gzip remained the default - but it exists, and it'd be nice for it to handle them, as well as helping the projects mentioned in #10.

cccs-sadugas:
> 
> 
> I'll re-review the changes, and perform benchmarks locally to confirm the performance impact.
> 
> From your results, it seems that there could be a slight performance regression even in non-streaming mode, maybe due to the changes in the range coder (`update` boolean parameter), in which case it could be worth putting that behind the `stream` feature as well to avoid the regression. If the regression is confirmed but the culprit is not the range coder, [producing a flamegraph](https://gendignoux.com/blog/2019/11/09/profiling-rust-docker-perf.html#profiling-rust-code) might be useful as well to see where the regression may come from.
> 
> I hope to get back with a more complete reply in the next few days.

@gendx any updates on this pull request?

I agree that the slight performance regression may be caused by the changes in the `process` function and the `update` flag as this creates additional code branches.

The first version of my pull request avoided this issue by writing two versions of the `process` and range coder functions at the cost of code duplication. E.g. `parse_reverse_bit_tree` with `&mut self` and `parse_reverse_bit_tree_check` with `&self`.

I ran the benchmarks once again today. This shows a larger **8-12%** difference for smaller tests and a lower **3%** difference for the big file test. Larger input may have better performance due to initial setup costs.

```
# master compared with this branch (default features)
 name                             master ns/iter  branch-default ns/iter  diff ns/iter  diff %  speedup 
 compress_65536                   1,889,807       1,806,011                    -83,796  -4.43%   x 1.05 
 compress_empty                   1,305           1,259                            -46  -3.52%   x 1.04 
 compress_hello                   1,707           1,710                              3   0.18%   x 1.00 
 decompress_after_compress_65536  1,695,077       1,896,378                    201,301  11.88%   x 0.89 
 decompress_after_compress_empty  3,010           3,074                             64   2.13%   x 0.98 
 decompress_after_compress_hello  3,540           3,817                            277   7.82%   x 0.93 
 decompress_big_file              5,950,045       6,104,193                    154,148   2.59%   x 0.97 
 decompress_huge_dict             3,670           3,672                              2   0.05%   x 1.00 
```

```
# this branch (default features) compared with all features
 name                             branch-default ns/iter  branch-all-features ns/iter  diff ns/iter  diff %  speedup 
 compress_65536                   1,806,011               1,846,704                          40,693   2.25%   x 0.98 
 compress_empty                   1,259                   1,309                                  50   3.97%   x 0.96 
 compress_hello                   1,710                   1,703                                  -7  -0.41%   x 1.00 
 decompress_after_compress_65536  1,896,378               1,874,422                         -21,956  -1.16%   x 1.01 
 decompress_after_compress_empty  3,074                   3,023                                 -51  -1.66%   x 1.02 
 decompress_after_compress_hello  3,817                   3,560                                -257  -6.73%   x 1.07 
 decompress_big_file              6,104,193               6,149,868                          45,675   0.75%   x 0.99 
 decompress_huge_dict             3,672                   3,647                                 -25  -0.68%   x 1.01 
WARNING: benchmarks in new but not in old: decompress_stream_big_file
```
```
# additional bench in this branch with all features
test decompress_stream_big_file      ... bench:   5,936,301 ns/iter (+/- 484,516)
```

I generated flamegraphs and attached them to this thread. N is the number of test iterations.
```
+-----------------+---------+----------------------------------+------+
|      name       | branch  |               test               |  N   |
+-----------------+---------+----------------------------------+------+
| master.svg      | master  | decompress_big_file foo.txt.lzma | 1000 |
| master-dac.svg  | master  | decompress_after_compress_65536  |  100 |
| control.svg     | current | decompress_big_file foo.txt.lzma | 1000 |
| control-dac.svg | current | decompress_after_compress_65536  |  100 |
+-----------------+---------+----------------------------------+------+
```

Flamegraph analysis:
- `diff.svg` shows the difference between master and control **+3%** overall mainly in `read`.
- `diff-dac.svg` shows the difference between master-dac and control-dac with **+40%** more time in `decode_bit`.

I tried doing some micro optimizations by moving branches but overall this did not create a significant performance increase. On the positive side, the streaming decompressor is more cpu and memory efficient :).

[flamegraph.zip](https://github.com/gendx/lzma-rs/files/5599627/flamegraph.zip)

gendx (approved):
Sorry again for the delay in reviewing this.

Overall this should be good to go as an experimental streaming feature. I left some minor comments for reference, which can be fixed separately after merging.

Regarding performance, the minor regression is not that large, at least when the feature is off. But further PRs are welcome to improve performance if needed.

However, I'd like to experiment with an `async` implementation in the future, e.g. over [futures::io::AsyncRead](https://docs.rs/futures/0.3.8/futures/io/trait.AsyncRead.html) and [futures::io::AsyncWrite](https://docs.rs/futures/0.3.8/futures/io/trait.AsyncWrite.html), which should provide a more seamless way of streaming without having to implement the state machine for buffering within lzma-rs, and in particular without having to have this `update` boolean passed throughout the implementation.

Other improvements could be to make `RangeDecoder` a trait and integrating the `update` bit into it (i.e. having `UpdatingRangeDecoder` and `ObservingRangeDecoder` that implement the `RangeDecoder` trait), or to leverage const generics to store the update bit as a const-generic boolean.

cccs-sadugas:
> 
> 
> Sorry again for the delay in reviewing this.
> 
> Overall this should be good to go as an experimental streaming feature. I left some minor comments for reference, which can be fixed separately after merging.
> 
> Regarding performance, the minor regression is not that large, at least when the feature is off. But further PRs are welcome to improve performance if needed.
> 
> However, I'd like to experiment with an `async` implementation in the future, e.g. over [futures::io::AsyncRead](https://docs.rs/futures/0.3.8/futures/io/trait.AsyncRead.html) and [futures::io::AsyncWrite](https://docs.rs/futures/0.3.8/futures/io/trait.AsyncWrite.html), which should provide a more seamless way of streaming without having to implement the state machine for buffering within lzma-rs, and in particular without having to have this `update` boolean passed throughout the implementation.
> 
> Other improvements could be to make `RangeDecoder` a trait and integrating the `update` bit into it (i.e. having `UpdatingRangeDecoder` and `ObservingRangeDecoder` that implement the `RangeDecoder` trait), or to leverage const generics to store the update bit as a const-generic boolean.

Thanks for the review and merge :). I will look at the comments in the near future. Is there anything else required to publish the added features to crates.io?

I agree that the async traits could be useful and is worth looking into. Some projects using this crate may not have an async runtime or may depend on using the `std::io::Write` interface. May be worth looking for a way to wrap the async call to turn it into an `std::io::Write` interface for those users.

gendx:
> Thanks for the review and merge :). I will look at the comments in the near future. Is there anything else required to publish the added features to crates.io?

Thanks also a lot for your patience, and sorry again for the reviewing delay, this was a big change and the last months have been quite busy on my end.
I'll give it a few more days before publishing a new version. In particular, it would be good to look at how this PR affects #61 before publishing anything.

> I agree that the async traits could be useful and is worth looking into. Some projects using this crate may not have an async runtime or may depend on using the `std::io::Write` interface. May be worth looking for a way to wrap the async call to turn it into an `std::io::Write` interface for those users.

The plan would be to have 3 implementations:
1. The main, blocking one.
1. The experimental streaming interface introduced here.
1. Another experimental `async` interface.

From a maintenance point-of-view, keeping only one of (2) and (3) would be preferable in the long term (depending on actual adoption), but on the other hand it's indeed not always possible/easy to have users pass an `AsyncWrite` object when this requires providing a fully-fledged async runtime (e.g. in embedded systems? in wasm?).

gendx:
> Thanks for the review and merge :). I will look at the comments in the near future. Is there anything else required to publish the added features to crates.io?

For now, the streaming implementation doesn't work for some combination(s) of the test cases and the streaming chunking size, as evidenced by #63, so fixing that is still required before publishing a new release on crates.io.

PR #61, opened 2020-12-02
Fix some major LZMA2 decompression issues

qnga:
### Pull Request Overview

This pull request fixes several issues with the LZMA2 decompression.
- Lzbuffer length reset. Although I could not find a file that this change allows to successfully decompress, I'm convinced it is needed. If I correctly understood, the Java implementation does the reset [there](https://github.com/sesuncedu/xz/blob/689d1b42e84adcbd5bbc78084a89f1e1d2d87e12/src/org/tukaani/xz/lz/LZDecoder.java#L46).
- Unpacked_size mismatches (for instance, it's likely to solve #11). The unpacked_size provided in the LZMA2 header is the size of the unpacked chunk, whereas lzbuffer.len() that was used before, is the size of the dict since the last reset. Not every chunk leads to a dict reset.
- LZMA2 header parsing. In some cases, properties were not reset while they should be.

### Testing Strategy

This pull request was tested by myself. It allows to decompress several files whose decompression failed before. This is the case of the _good-1-lzma2-4.xz_ file from the XZ test suite, so I added it in the test directory. The last two fixes are required to decompress it.

Note that this PR seems to improve the decompression of some other tests from the XZ test suite: decompression fails because of CRC errors instead of invalid status errors.

### TODO or Help Wanted

This pull request is based on the v0.1.3 version and still needs to be rebased on top of master. Lots of changes occurred since the v0.1.3 version and for some reason I'm not able to reproduce my results with master so far. I'm afraid a regression might have been introduced at some point. I'll try again but any help would be welcome.


gendx:
Thanks for taking a look at this!

> ### TODO or Help Wanted
> This pull request is based on the v0.1.3 version and still needs to be rebased on top of master. Lots of changes occurred since the v0.1.3 version and for some reason I'm not able to reproduce my results with master so far. I'm afraid a regression might have been introduced at some point. I'll try again but any help would be welcome.

I merged #59 this week, which refactored quite a bit of code. But even if your bug ended up being fixed in the meantime, adding more test cases is always welcome!

qnga:
That was the contrary, my bug hadn't been fixed in the meantime and I was not able to fix it on master. The good news is that I succeeded today, and three more test cases from the XZ test suite pass now with master + my own fixes.

gendx:
This overall looks good.

Just a couple comments regarding the failing Travis-CI:

- CI now fails for Rust 1.32.0, due to some dev dependency being now only compatible with Rust >= 1.34.0. I'm working on #62 to fix that.
- Formatting is incorrect, as shown in https://travis-ci.org/github/gendx/lzma-rs/jobs/747669312. Please run `cargo fmt`.

qnga:
Okay, I've reformatted the code according to cargo's rules.

PR #63, opened 2020-12-12
Prepare 0.2.0

gendx:
### Pull Request Overview

This pull request fixes various pending comments from #59 to prepare the release of version 0.2.0.

### Testing Strategy

This pull request was tested by Travis-CI

N/A

### TODO or Help Wanted

For now, streaming decompression fails in some test cases, which is blocking this release. CC @cccs-sadugas.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rust:
   - nightly
   - beta
   - stable
-  - 1.34.0  # pinned minimum supported toolchain
+  - 1.36.0  # pinned minimum supported toolchain
 
 matrix:
   allow_failures:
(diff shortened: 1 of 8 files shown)


cccs-sadugas:
Thanks for adding more chunk size testing. I'll have a look at the test results today.

cccs-sadugas:
Only chunks of size of `1` and `compressed.len()` work. All others do not: `2, 3, 4, 5, 6, 7, 8, 16, 32, 64, 128, 256, 512, 1024`.

However, this assertion holds: `assert_eq!(&decomp[..], &expected[..decomp.len()])`. There is an additional 599 bytes of all zeros in the expected value for all of the failing chunk sizes. More debugging to follow.

cccs-sadugas:
Finally found the bug. The `partial_input_buf` was not being accounted for in the eof checks. PR is coming shortly with a fuzzing target that helped a lot to find a smaller failing input. Sorry for the late response as I was busy releasing other open source projects.

PR #65, opened 2021-01-13
Fix Streaming Partial EOF Bug

cccs-sadugas:
### Pull Request Overview

This pull request fixes the bug in the streaming decompressor that produces shorter output than expected for some chunk sizes.

It also adds a streaming fuzz target.

### Testing Strategy

@gendx already introduced a test case for various chunk sizes which allowed us to find this bug. The fuzzing target will also help.

Related to https://github.com/gendx/lzma-rs/pull/63

### DIFF
--- a/fuzz/fuzz_targets/decompress_lzma_stream.rs
+++ b/fuzz/fuzz_targets/decompress_lzma_stream.rs
@@ -1,0 +1,40 @@
+#![no_main]
+#[macro_use]
+extern crate libfuzzer_sys;
+
+use lzma_rs::error::Result;
+use std::io::Write;
+
+fn decode_lzma(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut bf = std::io::Cursor::new(compressed);
+    let mut decomp: Vec<u8> = Vec::new();
+    lzma_rs::lzma_decompress(&mut bf, &mut decomp)?;
+    Ok(decomp)
+}
+
+fn decode_lzma_stream(compressed: &[u8], chunk_size: usize) -> Result<Vec<u8>> {
+    let mut stream = lzma_rs::decompress::Stream::new(Vec::new());
+    for chunk in compressed.chunks(chunk_size) {
+        stream.write_all(chunk).unwrap();
+    }
+    let decomp = stream.finish().unwrap();
+    Ok(decomp)
+}
+
+fuzz_target!(|input: &[u8]| {
+    let chunk_size = 2;
+    let mut input = std::io::Cursor::new(input);
+    let mut data = Vec::new();
+    lzma_rs::lzma_compress(&mut input, &mut data).unwrap();
+    let decomp_ref = decode_lzma(&data).unwrap();
+    let decomp_act = decode_lzma_stream(&data, chunk_size as usize).unwrap();
+    if decomp_act.len() != decomp_ref.len() {
+        panic!(
+            "chunk size: {}, ref len: {}, act len: {}",
+            chunk_size,
+            decomp_ref.len(),
+            decomp_act.len()
+        );
+    }
+    assert_eq!(decomp_act, decomp_ref);
+});
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -14,6 +14,7 @@ xz2 = "0.1.6"
 
 [dependencies.lzma-rs]
 path = ".."
+features = ["stream"]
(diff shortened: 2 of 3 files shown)


fuzz/fuzz_targets/decompress_lzma_stream.rs
```diff
@@ -0,0 +1,40 @@
+#![no_main]
+#[macro_use]
+extern crate libfuzzer_sys;
+
+use lzma_rs::error::Result;
+use std::io::Write;
+
+fn decode_lzma(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut bf = std::io::Cursor::new(compressed);
```
gendx, line 9: Nit: `bf` is not really a clear name. I'd suggest just importing `std::io::Cursor` in this file, and directly inline this without giving it a name, i.e. `lzma_decompress(&mut Cursor(compressed), ...)`.

fuzz/fuzz_targets/decompress_lzma_stream.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut stream = lzma_rs::decompress::Stream::new(Vec::new());
+    for chunk in compressed.chunks(chunk_size) {
+        stream.write_all(chunk).unwrap();
+    }
+    let decomp = stream.finish().unwrap();
+    Ok(decomp)
+}
+
+fuzz_target!(|input: &[u8]| {
+    let chunk_size = 2;
```
gendx, line 25: Here I think it would be interesting to fuzz various chunk sizes depending on the input. I'd suggest defining the chunk size as the first byte of the input, and let the rest of the input be the actual data. Something like the following:

```rust
let (chunk_size, data) = input.split_at(1);
let chunk_size = chunk_size[0] as usize;
```

This won't work when the input is empty, but in that case we can simply return directly.

fuzz/fuzz_targets/decompress_lzma_stream.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut input = std::io::Cursor::new(input);
+    let mut data = Vec::new();
```
gendx, line 27: This should be named "compressed", as "data" doesn't specify whether things are compressed or not.

fuzz/fuzz_targets/decompress_lzma_stream.rs
```diff
@@ -0,0 +1,40 @@
...
+    lzma_rs::lzma_compress(&mut input, &mut data).unwrap();
+    let decomp_ref = decode_lzma(&data).unwrap();
```
gendx, line 29: Please spell out full words, e.g. `decompressed` and `decompressed_streaming` (rather than `decomp_ref` and `decomp_act` respectively).

fuzz/fuzz_targets/decompress_lzma_stream.rs
```diff
@@ -0,0 +1,40 @@
...
+    let decomp_act = decode_lzma_stream(&data, chunk_size as usize).unwrap();
+    if decomp_act.len() != decomp_ref.len() {
+        panic!(
+            "chunk size: {}, ref len: {}, act len: {}",
+            chunk_size,
+            decomp_ref.len(),
+            decomp_act.len()
+        );
+    }
+    assert_eq!(decomp_act, decomp_ref);
```
gendx, line 39: I guess we should also assert that the decompressed data is equal to the original input.

gendx (commented):
Thanks for taking a look! I left some comments regarding style, other than that it looks good.

gendx:
If you can add the reduced test case that you found with fuzzing that would also be nice to have as an additional regression test.

fuzz/fuzz_targets/decompress_lzma_stream.rs
simdugas: I've added this which should be fine for smaller inputs. However, we could improve performance by removing the comparison to the non-streaming decompressed output.

cccs-sadugas:
Superseded by #67

cccs-sadugas:
> 
> 
> If you can add the reduced test case that you found with fuzzing that would also be nice to have as an additional regression test.

This PR is destined to merge into master, however that test would need #63 . How should I proceed?

gendx:
Ok, let me merge #67 first. I'll then update #63 again and hopefully merge it. We can add the reduced test case afterwards.

Issue #72, opened 2022-05-11
Expose decoder/encoder primitives as public


chyyran:
I maintain a [Rust implementation](https://github.com/SnowflakePowered/chd-rs) of MAME's [Compressed Hunks of Data](https://github.com/mamedev/mame/blob/master/src/lib/util/chd.h) format, which is essentially chunks of data compressed using various compression algorithms, LZMA included. One of the quirks of this format is that because they use a very old LZMA 19.0, the encoding parameters are not saved into the output stream, and therefore every CHD decoder has to essentially [mimic the defaults of LZMA 19.0](https://github.com/SnowflakePowered/chd-rs/blob/master/chd-rs/src/compression/lzma.rs#L65).

Thankfully lzma-rs allows this but since the `encode` and `decode` modules aren't public, I had to [fork and vendor](https://github.com/SnowflakePowered/lzma-rs) lzma-rs to access those primitives like `LzmaParams` in my crate. Additionally I added `LzmaParams::new` to construct an instance manually for this purpose.

Ideally I would like to just be able to use mainline `lzma-rs`. Since `decoder` is mostly documented already, would it be feasible to expose those primitives as public without much work, possibly behind a feature flag?

gendx:
> Since `decoder` is mostly documented already, would it be feasible to expose those primitives as public without much work, possibly behind a feature flag?

It surely would be feasible! Could you send a pull request for review?

I don't think a feature flag is needed here - just make sure that the newly exposed API is documented and that the GitHub workflows pass.

chyyran:
I did some internal refactoring to expose [the raw decoder as a primitive](https://docs.rs/lzma-rs-headerless/0.4.1-api-alpha/lzma_rs_headerless/decompress/struct.LzmaDecoder.html) where the LZMA compression parameters are encoded aas const generics. I also did a similar wrapper for [LZMA2](https://docs.rs/lzma-rs-headerless/0.4.1-api-alpha/lzma_rs_headerless/decompress/struct.Lzma2Decoder.html) although I'm not too sure about the usefulness of that in my fork. 

I believe this is a better approach than just exposing `LzmaParams` and I was able to expose this API without changing the existing public API, but at the cost of having to make some [relatively sizeable internal changes](https://github.com/SnowflakePowered/lzma-rs/commits/preserve-lzma-api), particularly with `DecoderState` to also allow reusing the existing allocation.

[...]

Issue #33, opened 2020-04-15
Test against the XZ test suite


Shnatsel:
The XZ utils source distribution at https://tukaani.org/xz/ contains 63 small files exercising various features of the format. The test files are placed in the public domain and there is a description of the expected behavior for each one. 

It would be nice to test lzma-rs against those files - it would have caught #32, for example.

Shnatsel:
See also #20

Shnatsel:
xz2 crate already does this, its test harness can be used as reference: https://github.com/alexcrichton/xz2-rs/blob/master/tests/xz.rs

gendx:
Note: these files can also be found in https://github.com/xz-mirror/xz/tree/master/tests/files.

Issue #20, opened 2019-12-11
Add tests against reference libraries


gendx:
To check that the implementation is correct, there should be tests against reference libraries such as https://crates.io/crates/lzma-sys, for example:
- compressing with one implementation and decompressing with the other,
- decompressing with both implementations and checking that the result (or error) is the same.

Shnatsel:
I had the same idea. Fuzzing can be of great help for discovering actually interesting inputs. This sounds fairly trivial, I'll see if I can whip up an implementation.

PR #73, opened 2022-06-01
Detach lifetime of output buffer from DecoderState

chyyran:
### Pull Request Overview
This pull request introduces **no new APIs** and makes no modifications to existing ones. There are no new features added. It is prep work for future features as per https://github.com/gendx/lzma-rs/issues/72.

This PR does two things.

* Removes ownership of `output` from `DecoderState`.
  * This removes the restriction of `DecoderState` to be parameterized on the lifetime of the output buffer to enable more flexible usage.
* Reuses already allocated buffers where possible for a small performance increase and additional flexibility for `DecoderState`
  * No public APIs have changed. Observable behaviour should be identical for existing consumers.

### Testing Strategy

- [x] The existing test suite is sufficient and should all pass. There should be no observable changes in behaviour or public API.

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -1,9 +1,8 @@
-use crate::decode::lzbuffer;
+use crate::decode::lzbuffer::LzBuffer;
 use crate::decode::rangecoder;
 use crate::error;
 use byteorder::{LittleEndian, ReadBytesExt};
 use std::io;
-use std::marker::PhantomData;
 
 use crate::decompress::Options;
 use crate::decompress::UnpackedSize;
@@ -38,14 +37,39 @@ enum ProcessingStatus {
     Finished,
 }
 
+#[derive(Copy, Clone)]
+/// LZMA 'lclppb' decompression properties.
+pub struct LzmaProperties {
+    /// The number of literal context bits.
+    ///
+    /// The most `lc` significant bits of the previous byte are part of the literal context.
+    /// `lc` must not be greater than 8.
+    pub lc: u32, // 0..=8
+    /// The number of literal position bits.
+    ///
+    /// `lp` must not be greater than 4.
+    pub lp: u32, // 0..=4
+    /// The number of position bits.
+    ///
+    /// The context for literal/match is plaintext offset modulo `2^pb`.
+    /// `pb` must not be greater than 4.
+    pub pb: u32, // 0..=4
+}
+
+impl LzmaProperties {
+    /// Assert the validity of the LZMA properties.
+    pub(crate) fn validate(&self) {
+        assert!(self.lc <= 8);
+        assert!(self.lp <= 4);
+        assert!(self.pb <= 4);
+    }
+}
+
+/// LZMA decompression parameters.
 pub struct LzmaParams {
-    // most lc significant bits of previous byte are part of the literal context
-    lc: u32, // 0..8
-    lp: u32, // 0..4
-    // context for literal/match is plaintext offset modulo 2^pb
-    pb: u32, // 0..4
-    dict_size: u32,
-    unpacked_size: Option<u64>,
+    pub properties: LzmaProperties,
+    pub dict_size: u32,
+    pub unpacked_size: Option<u64>,
 }
 
 impl LzmaParams {
@@ -108,9 +132,7 @@ impl LzmaParams {
         lzma_info!("Unpacked size: {:?}", unpacked_size);
 
         let params = LzmaParams {
-            lc,
-            lp,
-            pb,
+            properties: LzmaProperties { lc, lp, pb },
             dict_size,
             unpacked_size,
         };
@@ -119,21 +141,11 @@ impl LzmaParams {
     }
 }
 
-pub struct DecoderState<W, LZB>
-where
-    W: io::Write,
-    LZB: lzbuffer::LzBuffer<W>,
-{
-    _phantom: PhantomData<W>,
+pub(crate) struct DecoderState {
     // Buffer input data here if we need more for decompression. Up to
     // MAX_REQUIRED_INPUT bytes can be consumed during one iteration.
     partial_input_buf: std::io::Cursor<[u8; MAX_REQUIRED_INPUT]>,
-    pub output: LZB,
-    // most lc significant bits of previous byte are part of the literal context
-    pub lc: u32, // 0..8
-    pub lp: u32, // 0..4
-    // context for literal/match is plaintext offset modulo 2^pb
-    pub pb: u32, // 0..4
+    pub(crate) lzma_props: LzmaProperties,
     unpacked_size: Option<u64>,
     literal_probs: Vec<Vec<u16>>,
     pos_slot_decoder: Vec<rangecoder::BitTree>,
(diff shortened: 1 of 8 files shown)


src/decode/lzma.rs
```diff
@@ -188,33 +170,9 @@ where
...
-// Initialize decoder with circular buffer
-pub fn new_circular_with_memlimit<W>(
-    output: W,
-    params: LzmaParams,
-    memlimit: usize,
-) -> error::Result<DecoderState<W, lzbuffer::LzCircularBuffer<W>>>
-where
-    W: io::Write,
-{
+pub fn new_circular(params: LzmaParams) -> error::Result<DecoderState> {
```
gendx, line 173: With the `output` extracted out, this becomes identical to `new_accum`. Can you merge the two into a single function called `new`?

src/decode/lzma.rs
```diff
@@ -152,20 +145,9 @@ where
...
-    lc: u32,
-    lp: u32,
-    pb: u32,
-    unpacked_size: Option<u64>,
-) -> DecoderState<W, lzbuffer::LzAccumBuffer<W>>
-where
-    W: io::Write,
-{
+pub fn new_accum(lc: u32, lp: u32, pb: u32, unpacked_size: Option<u64>) -> DecoderState {
     DecoderState {
```
gendx, line 149: I think this was missed before, but can we add some assertions on `lc`, `lp` and `pb` to prevent invalid values? I.e. `lc < 8`, `lp < 4` and `pb < 4`?

src/decode/lzma.rs
```diff
@@ -239,48 +197,53 @@ where
...
 }
 
-impl<W, LZB> DecoderState<W, LZB>
-where
-    W: io::Write,
-    LZB: lzbuffer::LzBuffer<W>,
-{
+impl DecoderState {
     pub fn reset_state(&mut self, lc: u32, lp: u32, pb: u32) {
+        if self.lc + self.lp == lc + lp {
```
gendx, line 202: Likewise, the values of `lc`, `lp` and `pb` should be validated here.

src/decode/lzma.rs
```diff
@@ -239,48 +197,53 @@ where
...
+            // we can just reset here..
```
gendx, line 203: Nit: "We can just reset here."

src/decode/lzma.rs
```diff
@@ -239,48 +197,53 @@ where
...
+            self.literal_probs.iter_mut().for_each(|v| v.fill(0x400))
+        } else {
+            // need to reallocate because of new size of lc+lp
```
gendx, line 206: Nit: "We need to reallocate because of the new size of `lc+lp`."

src/decode/lzma.rs
```diff
@@ -239,48 +197,53 @@ where
...
         rangecoder: &mut rangecoder::RangeDecoder<'a, R>,
     ) -> error::Result<()> {
-        self.process_mode(rangecoder, ProcessingMode::Finish)
+        self.process_mode(output, rangecoder, ProcessingMode::Finish)
     }
 
     #[cfg(feature = "stream")]
     pub fn process_stream<'a, R: io::BufRead>(
         &mut self,
+        output: &mut LZB,
```
gendx, line 243: `LZB` isn't defined in this function.

src/decode/rangecoder.rs
```diff
@@ -180,6 +180,11 @@ impl BitTree {
     ) -> io::Result<u32> {
         rangecoder.parse_reverse_bit_tree(self.num_bits, self.probs.as_mut_slice(), 0, update)
     }
+
+    #[inline]
```
gendx, line 184: This shouldn't be necessary. If this really brings a performance improvement (assuming the compiler didn't already inline it by default), let's do that in a separate PR with proper before/after benchmarks.

src/decode/rangecoder.rs
```diff
@@ -215,4 +220,13 @@ impl LenDecoder {
             Ok(self.high_coder.parse(rangecoder, update)? as usize + 16)
         }
     }
+
+    #[inline]
```
gendx, line 224: Likewise here.

src/decode/stream.rs
```diff
@@ -155,24 +156,23 @@ where
...
             Ok(params) => {
-                let decoder = if let Some(memlimit) = options.memlimit {
-                    new_circular_with_memlimit(output, params, memlimit)
-                } else {
-                    new_circular(output, params)
-                }?;
-
+                let dict_size = params.dict_size;
+                let decoder = new_circular(params)?;
+                let output =LzCircularBuffer::from_stream_with_memlimit(output, dict_size as usize,
```
gendx, line 161: Is `LzCircularBuffer::from_stream` still used after this change? If not, let's just remove it and rename `from_stream_with_memlimit` to `from_stream`.

gendx:
I've approved the workflows to run. As there are some formatting errors, could you fix them as well and squash your changes into a single commit (force-pushed to this branch), to have a clean git history?

src/decode/lzma.rs
```diff
@@ -239,48 +197,53 @@ where
...
-impl<W, LZB> DecoderState<W, LZB>
-where
-    W: io::Write,
-    LZB: lzbuffer::LzBuffer<W>,
-{
+impl DecoderState {
     pub fn reset_state(&mut self, lc: u32, lp: u32, pb: u32) {
+        if self.lc + self.lp == lc + lp {
+            // we can just reset here..
+            self.literal_probs.iter_mut().for_each(|v| v.fill(0x400))
```
gendx, line 204: This requires Rust 1.50+. Can you increase the MSRV as such in https://github.com/gendx/lzma-rs/blob/master/.github/workflows/tests.yml#L14 and https://github.com/gendx/lzma-rs/blob/master/README.md?

gendx (changes requested):

src/decode/lzma.rs
chyyran: I've renamed this to `new` and moved it to an associated function of `DecoderState`

src/decode/lzma.rs
chyyran: Done:

https://github.com/gendx/lzma-rs/pull/73/files#diff-768757151baad4fb0605aaeee5210dc343977125d64711ebcf5a0e755b2f23c0R177

src/decode/lzma.rs
chyyran: https://github.com/gendx/lzma-rs/pull/73/files#diff-768757151baad4fb0605aaeee5210dc343977125d64711ebcf5a0e755b2f23c0R150

src/decode/stream.rs
chyyran: Renamed `LzCircularBuffer::from_stream_with_memlimit ` to `from_stream`. Also renamed `LzAccumBuffer::from_stream_with_memlimit` to `from_stream` to mirror the `LzCircularBuffer` API. The only consumer of `LzAccumBuffer::from_stream` was `lzma2`, where I just passed `usize::MAX` as the default limit.

src/decode/lzma.rs
```diff
@@ -151,136 +144,86 @@ where
...
-        is_rep_g1: [0x400; 12],
-        is_rep_g2: [0x400; 12],
-        is_rep_0long: [0x400; 192],
-        state: 0,
-        rep: [0; 4],
-        len_decoder: rangecoder::LenDecoder::new(),
-        rep_len_decoder: rangecoder::LenDecoder::new(),
+impl DecoderState {
+    // Initialize decoder with accumulating buffer
+    pub fn new(lc: u32, lp: u32, pb: u32, unpacked_size: Option<u64>) -> Self {
```
gendx, line 149: Rather than passing the parameters one by one to the function, could this API use `LzmaParams` struct? This would avoid potential mistakes in swapping `lc`/`lp`/`pb` between the call site and the function definition. Or at the very least define a new `LzmaProperties` struct for `(lc, lp, pb)`, and use that for all functions that take as input these 3 parameters together.

src/decode/lzma.rs
```diff
@@ -151,136 +144,86 @@ where
...
-) -> error::Result<DecoderState<W, lzbuffer::LzCircularBuffer<W>>>
-where
-    W: io::Write,
-{
-    new_circular_with_memlimit(output, params, std::usize::MAX)
-}
+    pub fn reset_state(&mut self, lc: u32, lp: u32, pb: u32) {
+        assert!(lc < 8);
+        assert!(lp < 4);
+        assert!(pb < 4);
```
gendx, line 180: I now realized that the ranges in the documentation are inconsistent with their actual definition in `LzmaParams` above:

```rust
        let lc: u32 = pb % 9;
        pb /= 9;
        let lp: u32 = pb % 5;
        pb /= 5;
```

Can you update the assertions, and also the docs to e.g. `0..=8`?

Ideally, the boundary condition on these values is also something we'd want to test (either as part of #33, or as new test cases - and also comparing to other implementations i.e. #20).

gendx (changes requested):

src/decode/lzma.rs
chyyran: Fixed this and moved assertions to `LzmaProperties::validate`

src/decode/lzma.rs
chyyran: For now I've defined an `LzmaProperties` struct for `lclppb`. There may be a cleaner way of exposing this but this can be put off for bikeshedding a public raw decoder API.

src/decode/lzma2.rs
```diff
@@ -138,24 +147,24 @@ where
 
             lzma_info!("Properties {{ lc: {}, lp: {}, pb: {} }}", lc, lp, pb);
         } else {
-            lc = decoder.lc;
-            lp = decoder.lp;
-            pb = decoder.pb;
+            lc = decoder.lzma_props.lc;
+            lp = decoder.lzma_props.lp;
+            pb = decoder.lzma_props.pb;
```
gendx, line 152: Nit: This block could directly return an `LzmaProperties` struct, rather than extracting the properties here to re-merge them just afterwards.

src/decode/lzma.rs
```diff
@@ -38,14 +37,39 @@ enum ProcessingStatus {
     Finished,
 }
 
-pub struct LzmaParams {
+#[derive(Copy, Clone)]
+/// LZMA 'lclppb' decompression properties.
+pub struct LzmaProperties {
     // most lc significant bits of previous byte are part of the literal context
```
gendx, line 43: Nit: This explanation can probably be merged with the following line, i.e.:

```
/// The number of literal context bits.
///
/// The most `lc` significant bits of the previous byte are part of the literal context.
```

src/decode/lzma.rs
```diff
@@ -38,14 +37,39 @@ enum ProcessingStatus {
...
-    lp: u32, // 0..4
+    /// The number of literal context bits.
+    ///
+    /// `lc` must not be greater than 8.
+    pub lc: u32, // 0..=8
+    /// The number of literal position bits.
+    ///
+    /// `lp` must not be greater than 4.
+    pub lp: u32, // 0..=4
     // context for literal/match is plaintext offset modulo 2^pb
```
gendx, line 52: Likewise:

```
/// The number of position bits.
///
/// The context for literal/match is the plaintext offset modulo `2^pb`.
```

src/decode/lzma.rs
```diff
@@ -151,136 +163,76 @@ where
...
-        is_rep_g0: [0x400; 12],
-        is_rep_g1: [0x400; 12],
-        is_rep_g2: [0x400; 12],
-        is_rep_0long: [0x400; 192],
-        state: 0,
-        rep: [0; 4],
-        len_decoder: rangecoder::LenDecoder::new(),
-        rep_len_decoder: rangecoder::LenDecoder::new(),
+impl DecoderState {
+    // Initialize decoder with accumulating buffer
```
gendx, line 167: There is no accumulating buffer anymore here. IMO the function is self-explanatory, the comment can be removed.

gendx (changes requested):

src/decode/lzma.rs
chyyran: https://github.com/gendx/lzma-rs/pull/73/commits/5716a6bca5872c33ff806ccfd6a87b6ea90e31b9#diff-768757151baad4fb0605aaeee5210dc343977125d64711ebcf5a0e755b2f23c0R167

src/decode/lzma.rs
chyyran: https://github.com/gendx/lzma-rs/pull/73/commits/5716a6bca5872c33ff806ccfd6a87b6ea90e31b9#diff-768757151baad4fb0605aaeee5210dc343977125d64711ebcf5a0e755b2f23c0R45

src/decode/lzma2.rs
chyyran: https://github.com/gendx/lzma-rs/pull/73/commits/5716a6bca5872c33ff806ccfd6a87b6ea90e31b9#diff-aec280556b32db2cd7e0d9b67090c0833582fa5e5dac272202f84adcfd64e887R119

Moved the in-flight lc/lp/pb values inside of the `reset_props` blocks as well.

src/decode/lzma2.rs
```diff
@@ -103,15 +112,15 @@ where
...
     }
 
     if reset_state {
-        let lc: u32;
-        let lp: u32;
-        let mut pb: u32;
+        let new_props = if reset_props {
+            let lc: u32;
+            let lp: u32;
+            let mut pb: u32;
```
gendx, line 122: Please check and fix the lints in this function as reported by Clippy: https://github.com/gendx/lzma-rs/runs/6790343456

gendx (changes requested):

gendx:
Overall this looks very good! There are just some lint failures left, but you can ignore the ones in `src/decode/rangecoder.rs` as they are unrelated to this change (probably some new lints enabled recently, I'll make a separate PR to fix them).

src/decode/lzma2.rs
chyyran: fixed: https://github.com/gendx/lzma-rs/commit/163b034954e2c0295a8e25c5567559fdccce2470#diff-aec280556b32db2cd7e0d9b67090c0833582fa5e5dac272202f84adcfd64e887R124

PR #74, opened 2022-06-09
Expose API for raw decoder primitives for LZMA and LZMA2.

### LINKED ISSUE #72: Expose decoder/encoder primitives as public

chyyran:
### Pull Request Overview

This pull request fixes #72 by refactoring decoder functions to expose a zero-cost raw decoder API. The existing public API has been refactored to use the raw decoder API internally. The following additions to the public API become available with the `raw_decoder` feature enabled under the `decompress::raw` module.

* `LzmaDecoder`
  * `LzmaDecoder::new(params: LzmaParams, memlimit: Option<usize>) -> Result<Self>`
  * `LzmaDecoder::reset(&mut self, unpacked_size: Option<Option<u64>>)`
  * `LzmaDecoder::decompress<'a, W: io::Write, R: io::BufRead>(&mut self, input: &mut R, output: &'a mut W)`
* `Lzma2Decoder`
  * `Lzma2Decoder::new() -> Self`
  * `Lzma2Decoder::reset(&mut self)`
  * `Lzma2Decoder::decompress<'a, W: io::Write, R: io::BufRead>(&mut self, input: &mut R, output: &'a mut W)`
* `LzmaProperties`
* `LzmaParams`
  * `LzmaParams::new(properties: LzmaProperties, dict_size: u32, unpacked_size: Option<u64>) -> Self`
  * `LzmaParams:read_header<R>(input: &mut R, options: &Options) -> error::Result<LzmaParams>`
     * This was a previously internal API that is now exposed only when `raw_decoder` is enabled. I don't see any reason to restrict this to internal usage.

Additionally, annotations have been added to indicate availability of APIs on docs.rs.
image

### Testing Strategy
- [x] The existing test suite is sufficient and should all pass. The existing API tests have been refactored to use the `raw_decoder` APIs internally.

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -1,11 +1,11 @@
-use crate::decode::lzbuffer::LzBuffer;
+use crate::decode::lzbuffer::{LzBuffer, LzCircularBuffer};
 use crate::decode::rangecoder;
+use crate::decode::rangecoder::RangeDecoder;
+use crate::decompress::Options;
+use crate::decompress::UnpackedSize;
 use crate::error;
 use byteorder::{LittleEndian, ReadBytesExt};
 use std::io;
-
-use crate::decompress::Options;
-use crate::decompress::UnpackedSize;
 
 /// Maximum input data that can be processed in one iteration.
 /// Libhtp uses the following equation to define the maximum number of bits
@@ -37,7 +37,7 @@ enum ProcessingStatus {
     Finished,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 /// LZMA 'lclppb' decompression properties.
 pub struct LzmaProperties {
     /// The number of literal context bits.
@@ -65,14 +65,33 @@ impl LzmaProperties {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
 /// LZMA decompression parameters.
 pub struct LzmaParams {
-    pub properties: LzmaProperties,
-    pub dict_size: u32,
-    pub unpacked_size: Option<u64>,
+    /// The LZMA 'lclppb' decompression properties.
+    pub(crate) properties: LzmaProperties,
+    /// The dictionary size to use when decompressing.
+    pub(crate) dict_size: u32,
+    /// The size of the unpacked data.
+    pub(crate) unpacked_size: Option<u64>,
 }
 
 impl LzmaParams {
+    /// Create an new instance of LZMA parameters.
+    #[cfg(feature = "raw_decoder")]
+    pub fn new(
+        properties: LzmaProperties,
+        dict_size: u32,
+        unpacked_size: Option<u64>,
+    ) -> LzmaParams {
+        Self {
+            properties,
+            dict_size,
+            unpacked_size,
+        }
+    }
+
+    /// Read LZMA parameters from the LZMA stream header.
     pub fn read_header<R>(input: &mut R, options: &Options) -> error::Result<LzmaParams>
     where
         R: io::BufRead,
@@ -141,6 +160,7 @@ impl LzmaParams {
     }
 }
 
+#[derive(Debug)]
 pub(crate) struct DecoderState {
     // Buffer input data here if we need more for decompression. Up to
     // MAX_REQUIRED_INPUT bytes can be consumed during one iteration.
@@ -557,5 +577,56 @@ impl DecoderState {
         }
 
         Ok(result)
+    }
+}
+
+#[derive(Debug)]
+/// Raw decoder for LZMA.
+pub struct LzmaDecoder {
+    params: LzmaParams,
+    memlimit: usize,
+    state: DecoderState,
+}
+
+impl LzmaDecoder {
+    /// Creates a new object ready for decompressing data that it's given for the input
+    /// dict size, expected unpacked data size, and memory limit for the internal buffer.
+    pub fn new(params: LzmaParams, memlimit: Option<usize>) -> error::Result<LzmaDecoder> {
+        Ok(Self {
+            params,
+            memlimit: memlimit.unwrap_or(usize::MAX),
+            state: DecoderState::new(params.properties, params.unpacked_size),
+        })
+    }
+
+    /// Performs the equivalent of replacing this decompression state with a freshly allocated copy.
+    ///
+    /// Because the decoder state is reset, the unpacked size may optionally be re-specified. If `None`
+    /// is given, the previous unpacked size that the decoder was initialized with remains unchanged.
+    ///
+    /// This function may not allocate memory and will attempt to reuse any previously allocated resources.
+    #[cfg(feature = "raw_decoder")]
+    pub fn reset(&mut self, unpacked_size: Option<Option<u64>>) {
+        self.state.reset_state(self.params.properties);
+
+        if let Some(unpacked_size) = unpacked_size {
+            self.state.set_unpacked_size(unpacked_size);
+        }
+    }
+
+    /// Decompresses the input data into the output, consuming only as much input as needed and writing as much output as possible.
+    pub fn decompress<W: io::Write, R: io::BufRead>(
+        &mut self,
+        input: &mut R,
+        output: &mut W,
+    ) -> error::Result<()> {
+        let mut output =
+            LzCircularBuffer::from_stream(output, self.params.dict_size as usize, self.memlimit);
+
+        let mut rangecoder = RangeDecoder::new(input)
+            .map_err(|e| error::Error::LzmaError(format!("LZMA stream too short: {}", e)))?;
+        self.state.process(&mut output, &mut rangecoder)?;
+        output.finish()?;
+        Ok(())
     }
 }
--- a/src/decode/lzma2.rs
+++ b/src/decode/lzma2.rs
@@ -1,6 +1,6 @@
 use crate::decode::lzbuffer;
 use crate::decode::lzbuffer::LzBuffer;
-use crate::decode::lzma;
+use crate::decode::lzma::DecoderState;
 use crate::decode::lzma::LzmaProperties;
 use crate::decode::rangecoder;
 use crate::error;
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -44,17 +50,9 @@ pub fn lzma_decompress_with_options<R: io::BufRead, W: io::Write>(
...
-        options.memlimit.unwrap_or(usize::MAX),
-    );
-
-    let mut rangecoder = decode::rangecoder::RangeDecoder::new(input)
-        .map_err(|e| error::Error::LzmaError(format!("LZMA stream too short: {}", e)))?;
-    decoder.process(&mut output, &mut rangecoder)?;
-    output.finish()?;
+    let mut decoder = decode::lzma::LzmaDecoder::new(params, options.memlimit)?;
+
+    decoder.decompress(input, output)?;
```
chyyran, line 55: Decoding logic has been moved into `LzmaDecoder::decompress`.

chyyran:
The current API for `decompress` is very simple and matches up well with the existing one-shot functions. One downside is such a design is that the `LzBuffer`s are still one-shot and result in an allocation every time `decompress` is called. 

It might be worth looking into allowing reusage of a backing buffer in an `LzBuffer` but this requires further design work on the `LzBuffer` implementations that isn't really related to #72. The current API design does not preclude a future API that would allow reusage of these buffers.

chyyran:
cc @gendx

chyyran:
I encountered a situation where simply resetting the decoder state was not enough but since the decoder is being reused for each raw LZMA chunk, the unpacked size was also needed to be set. Since this might not be the case for every usage, I have it in `reset` as `Option<Option<u64>>` where `None` leaves the unpacked size unchanged, and `Some(_)` changes the unpacked size in the decoder state.

I thought about exposing `set_unpacked_size` to `LzmaDecoder` but it didn't really make sense in my perspective since `reset` prepares the decoder for the next chunk of data. For some streams, forgetting to call `set_unpacked_size` may leave the decoder in an invalid state *for that specific stream*, so keeping it atomic was best. I don't see any situation where someone would need to change the unpacked size between raw chunks.

src/decode/lzma.rs
```diff
@@ -559,3 +580,54 @@ impl DecoderState {
...
+        if let Some(unpacked_size) = unpacked_size {
+            self.decoder.set_unpacked_size(unpacked_size);
+        }
+    }
+
+    /// Decompresses the input data into the output, consuming only as much input as needed and writing as much output as possible.
+    pub fn decompress<'a, W: io::Write, R: io::BufRead>(
+        &mut self,
+        input: &mut R,
+        output: &'a mut W,
```
gendx, line 622: Is this explicit `'a` lifetime needed here? I don't see it anywhere else in this function so I guess it can be removed as per the [lifetime elision rules](https://doc.rust-lang.org/reference/lifetime-elision.html)?

src/decode/lzma.rs
```diff
@@ -559,3 +580,54 @@ impl DecoderState {
         Ok(result)
     }
 }
+
+#[derive(Debug)]
+/// Raw decoder for LZMA.
+pub struct LzmaDecoder {
+    params: LzmaParams,
+    memlimit: usize,
+    decoder: DecoderState,
```
gendx, line 589: Nit: a "decoder" within a decoder sounds weird, it might be worth calling this variable "state" (or "decoder_state") instead?

src/decode/lzma2.rs
```diff
@@ -193,3 +190,46 @@ where
 
     Ok(())
 }
+
+#[derive(Debug)]
+/// Raw decoder for LZMA2.
+pub struct Lzma2Decoder {
+    decoder: DecoderState,
```
gendx, line 197: Likewise, in terms of naming this would be better called `lzma_decoder`, to show that the LZMA2 encapsulates a raw LZMA decoder.

src/decode/lzma2.rs
```diff
@@ -1,27 +1,24 @@
...
+use crate::decode::lzma::DecoderState;
 use crate::decode::lzma::LzmaProperties;
 use crate::decode::rangecoder;
 use crate::error;
 use byteorder::{BigEndian, ReadBytesExt};
 use std::io;
 use std::io::Read;
 
-pub fn decode_stream<R, W>(input: &mut R, output: &mut W) -> error::Result<()>
+fn decode_stream<R, W>(
```
gendx, line 12: Nit: it might be worth moving these free functions into the `impl` block of the new `Lzma2Decoder` struct?

src/decode/xz.rs
```diff
@@ -352,7 +352,7 @@ where
                 )));
             }
             // TODO: properties??
-            lzma2::decode_stream(&mut count_input, output)?;
+            lzma2::Lzma2Decoder::new().decompress(&mut count_input, output)?;
```
gendx, line 355: Nit: now that there is a new struct, it'd be more consistent to import `lzma2::Lzma2Decoder`.

src/lib.rs
```diff
@@ -25,6 +24,13 @@ pub mod compress {
...
     pub use crate::decode::options::*;
+    #[cfg(feature = "raw_decoder")]
+    #[cfg_attr(docsrs, doc(cfg(raw_decoder)))]
+    pub mod raw {
+        //! Raw decoding primitives for LZMA/LZMA2 streams.
+        pub use crate::decode::lzma::{LzmaDecoder, LzmaParams, LzmaProperties};
+        pub use crate::decode::lzma2::Lzma2Decoder;
+    }
     #[cfg(feature = "stream")]
     pub use crate::decode::stream::Stream;
```
gendx, line 35: Would this also benefit from a `cfg_attr` on `docrs` and the `stream` feature?

src/lib.rs
```diff
@@ -44,17 +50,9 @@ pub fn lzma_decompress_with_options<R: io::BufRead, W: io::Write>(
...
-    );
-
-    let mut rangecoder = decode::rangecoder::RangeDecoder::new(input)
-        .map_err(|e| error::Error::LzmaError(format!("LZMA stream too short: {}", e)))?;
-    decoder.process(&mut output, &mut rangecoder)?;
-    output.finish()?;
+    let mut decoder = decode::lzma::LzmaDecoder::new(params, options.memlimit)?;
+
+    decoder.decompress(input, output)?;
     Ok(())
```
gendx, line 56: This can be simplified by having the function directly return the result from `decompress()`.

src/lib.rs
```diff
@@ -44,17 +50,9 @@ pub fn lzma_decompress_with_options<R: io::BufRead, W: io::Write>(
...
-        params.dict_size as usize,
-        options.memlimit.unwrap_or(usize::MAX),
-    );
-
-    let mut rangecoder = decode::rangecoder::RangeDecoder::new(input)
-        .map_err(|e| error::Error::LzmaError(format!("LZMA stream too short: {}", e)))?;
-    decoder.process(&mut output, &mut rangecoder)?;
-    output.finish()?;
+    let mut decoder = decode::lzma::LzmaDecoder::new(params, options.memlimit)?;
+
```
gendx, line 54: Nit: I'd remove this empty line given how short the function is now.

gendx (commented):
Sorry for the delay, I just came back from holidays.

Overall this looks good, I just left a few nits.
Thank you a lot for contributing this!

src/decode/lzma2.rs
chyyran: Since the one in `LzmaDecoder` is `state` how is `lzma_state` to be more consistent?

src/decode/lzma2.rs
chyyran: The only thing is the diff is a little more hard to read now but that doesn't really matter anymore.

chyyran: I moved `lzma2::decode_stream` directly into `Lzma2Decoder::decompress` and just made it take `&mut self`.

src/decode/lzma.rs
```diff
@@ -1,12 +1,13 @@
-use crate::decode::lzbuffer::LzBuffer;
+use crate::decode::lzbuffer::{LzBuffer, LzCircularBuffer};
+use crate::decode::rangecoder::RangeDecoder;
+use crate::decompress::Options;
+use crate::decompress::UnpackedSize;
+use byteorder::{LittleEndian, ReadBytesExt};
+
```
gendx, line 6: Formatting: please remove this empty line and re-run rustfmt.

src/decode/lzma2.rs
```diff
@@ -1,108 +1,75 @@
...
-        .map_err(|e| error::Error::LzmaError(format!("LZMA2 expected packed size: {}", e)))?;
-    let packed_size = (packed_size as u64) + 1;
+        let packed_size = input
+            .read_u16::<BigEndian>()
+            .map_err(|e| error::Error::LzmaError(format!("LZMA2 expected packed size: {}", e)))?;
+        let packed_size = (packed_size as u64) + 1;
 
-    lzma_info!(
+        lzma_info!(
         "LZMA2 compressed block {{ unpacked_size: {}, packed_size: {}, reset_dict: {}, reset_state: {}, reset_props: {} }}",
```
gendx, line 73: Formatting: please indent the macro's parameters.

src/decode/lzma2.rs
```diff
@@ -111,85 +78,146 @@ where
...
-        .map_err(|e| error::Error::LzmaError(format!("LZMA2 expected unpacked size: {}", e)))?;
-    let unpacked_size = (unpacked_size as usize) + 1;
-
-    lzma_info!(
-        "LZMA2 uncompressed block {{ unpacked_size: {}, reset_dict: {} }}",
-        unpacked_size,
-        reset_dict
-    );
+    /// Creates a new object ready for decompressing data that it's given.
+    pub fn new() -> Lzma2Decoder {
```
gendx, line 166: Nit: This constructor would make more sense at the beginning of the impl block, to be easier to find.

src/decode/lzma2.rs
```diff
@@ -111,85 +78,146 @@ where
...
+        }
+    }
 
-    if reset_dict {
-        accum.reset()?;
+    /// Performs the equivalent of replacing this decompression state with a freshly allocated copy.
+    ///
+    /// This function may not allocate memory and will attempt to reuse any previously allocated resources.
+    #[cfg(feature = "raw_decoder")]
+    pub fn reset(&mut self) {
```
gendx, line 183: Likewise this small public function should be at the top after the `new()`.

src/decode/lzma2.rs
```diff
@@ -111,85 +78,146 @@ where
...
-    let mut buf = vec![0; unpacked_size];
-    input.read_exact(buf.as_mut_slice()).map_err(|e| {
-        error::Error::LzmaError(format!(
-            "LZMA2 expected {} uncompressed bytes: {}",
-            unpacked_size, e
-        ))
-    })?;
-    accum.append_bytes(buf.as_slice());
+    /// Decompresses the input data into the output, consuming only as much input as needed and writing as much output as possible.
+    pub fn decompress<W: io::Write, R: io::BufRead>(
```
gendx, line 192: And as this one is public I'd put it at the top as well. This may also make the diff easier for GitHub.

gendx (approved):

jtmoon79:
Hi @chyyran @gendx

I wanted to read an XZ file into decompressed chunks of some constant size. Currently I'm using [xz_decompress](https://docs.rs/lzma-rs/0.3.0/lzma_rs/fn.xz_decompress.html). But `xz_decompress` reads the entire file in one call.  I want to call something that sequentially reads the XZ into some fixed size buffer (https://github.com/jtmoon79/super-speedy-syslog-searcher/issues/182).

Can I do read an XZ file as a sequence of decompressed byte chunks (`Vec<u8>`) using these new API endpoints?

Thanks for creating this library! 😄

Issue #19, opened 2019-12-10
Use const-generics


gendx:
For now, the `DecoderState` and `BitTree` structs use memory allocations (in `Vec`) to store their state, despite having sizes known at compile time. This is in the most part to avoid code duplication.

Once [const generics](https://github.com/rust-lang/rust/issues/44580) are stable enough, they should be used instead to remove the need for dynamic allocation, and make the whole state struct more compact and contiguous (therefore potentially more cache-friendly).

This should hopefully improve performance a bit.

npmccallum:
Const generics are now stable in 1.51.

gendx:
I had a brief look, the [`min_const_generics` feature](https://github.com/rust-lang/rust/issues/74878) was indeed stabilized in 1.51, but we still need the more complex [`const_evaluatable_checked` feature](https://blog.rust-lang.org/2021/02/26/const-generics-mvp-beta.html#const-generics-with-complex-expressions) due to the `1 << num_bits` expression [here](https://github.com/gendx/lzma-rs/blob/v0.2.0/src/decode/rangecoder.rs#L164).

In the meantime, it might be worth experimenting with it on nightly to see if benchmarks would improve with const generics.

chyyran:
I've been doing some experiments in https://github.com/chyyran/lzma-rs/commits/feature-perf-experiments and found that const generics in `BitTree` and `RangeCoder` doesn't actually contribute that much performance benefit, *except for large dictionaries*.

https://github.com/chyyran/lzma-rs/commit/fba32253c8fdca54b10379caa092d2d6a17d6984
```
running 8 tests
test compress_65536                  ... bench:   1,359,753 ns/iter (+/- 94,903)
test compress_empty                  ... bench:         701 ns/iter (+/- 45)
test compress_hello                  ... bench:       1,011 ns/iter (+/- 208)
test decompress_after_compress_65536 ... bench:   2,445,491 ns/iter (+/- 93,723)
test decompress_after_compress_empty ... bench:       1,841 ns/iter (+/- 93)
test decompress_after_compress_hello ... bench:       2,345 ns/iter (+/- 102)
test decompress_big_file             ... bench:   3,864,228 ns/iter (+/- 257,509)
test decompress_huge_dict            ... bench:       2,354 ns/iter (+/- 182)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out; finished in 18.29s
```

[...]

PR #77, opened 2022-08-06
Reduce unnecessary Vec allocations and indirections

chyyran:
### Pull Request Overview
* Changed literal_probs array from a `Vec<Vec<u16>>` to a Vec2D backed by a contiguous `Vec<u16>`.
* BitTrees in LenDecoder and DecoderState are now stored inline as an array. The actual BitTree data is still on the heap.
* DecoderState::reset now just recreates stack-allocated data, dropping the existing ones. I found this to be faster than using `slice::fill`. The heap-allocated arrays still use `fill` to keep their allocation.

This gives a small but free performance boost and reduces overall allocations.

https://github.com/gendx/lzma-rs/commit/6e1f0d797230ea3c30f9c002967f1151ac555124
```
running 8 tests
test compress_65536                  ... bench:   1,406,069 ns/iter (+/- 60,172)
test compress_empty                  ... bench:         734 ns/iter (+/- 36)
test compress_hello                  ... bench:       1,037 ns/iter (+/- 31)
test decompress_after_compress_65536 ... bench:   2,708,511 ns/iter (+/- 298,490)
test decompress_after_compress_empty ... bench:       2,243 ns/iter (+/- 1,853)
test decompress_after_compress_hello ... bench:       2,869 ns/iter (+/- 680)
test decompress_big_file             ... bench:   4,097,344 ns/iter (+/- 182,653)
test decompress_huge_dict            ... bench:       2,791 ns/iter (+/- 183)
```

After this PR (notice the tighter variance in particular)

```
running 8 tests
test compress_65536                  ... bench:   1,391,247 ns/iter (+/- 38,388)
test compress_empty                  ... bench:         725 ns/iter (+/- 17)
test compress_hello                  ... bench:       1,035 ns/iter (+/- 36)
test decompress_after_compress_65536 ... bench:   2,441,367 ns/iter (+/- 53,371)
test decompress_after_compress_empty ... bench:       1,974 ns/iter (+/- 66)
test decompress_after_compress_hello ... bench:       2,459 ns/iter (+/- 291)
test decompress_big_file             ... bench:   3,849,870 ns/iter (+/- 192,785)
test decompress_huge_dict            ... bench:       2,427 ns/iter (+/- 126)
```

### Testing Strategy


chyyran:
A previous version of this PR used [array-macro](https://crates.io/crates/array-macro) to initialize the BitTree arrays without needing to duplicate code. However, array-macro 2.0 has an MSRV of 1.56, and array-macro 1.0 uses `unsafe`, so I didn't think it was worth either of those tradeoffs to save a few lines of code.

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Debug)]
+pub struct Vec2D<T> {
+    data: Vec<T>,
+    cols: usize,
+    rows: usize,
+}
+
+impl<T> Vec2D<T> {
+    /// Initialize a grid of size `rows * cols` with the given data element.
+    pub fn init(rows: usize, cols: usize, data: T) -> Vec2D<T>
```
gendx, line 13: It'd be more logical to have the data as first parameter, to be consistent with the `vec![data; size]` style. Also, grouping together the dimensions as a tuple may make it more obvious that it's a size parameter:

`init(data: T, size: (usize, usize))`

src/decode/lzma.rs
```diff
@@ -1,6 +1,7 @@
 use crate::decode::lzbuffer::{LzBuffer, LzCircularBuffer};
 use crate::decode::rangecoder;
 use crate::decode::rangecoder::RangeDecoder;
+use crate::decode::vec2d::Vec2D;
```
gendx, line 4: It's not really a decode-specific code, so I'd put it in a `crate::util::` namespace.

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+        T: Clone,
+    {
+        self.data.fill(value)
+    }
+}
+
+impl<T> Index<usize> for Vec2D<T> {
+    type Output = [T];
+
+    fn index(&self, idx: usize) -> &Self::Output {
```
gendx, line 36: Instead of `idx`, can you name this variable `row` to make it more obvious to the reader?

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+            &self.data[start_idx..start_idx + self.cols]
+        } else {
+            panic!("row index {:?} out of bounds.", idx);
+        }
+    }
+}
+
+impl<T> IndexMut<usize> for Vec2D<T> {
+    fn index_mut(&mut self, idx: usize) -> &mut Self::Output {
+        &mut self.data[(idx * self.cols)..]
```
gendx, line 48: Why is this implementation different than `index`? As I understand it, this means `&mut vec2d[1][4]` would not panic.

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl<T> Vec2D<T> {
+    /// Initialize a grid of size `rows * cols` with the given data element.
+    pub fn init(rows: usize, cols: usize, data: T) -> Vec2D<T>
+    where
+        T: Clone,
+    {
+        Vec2D {
+            data: vec![data; rows * cols],
```
gendx, line 18: This `rows * cols` may overflow. Can you use a checked multiplication and panic in case of overflow?

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+        vec2d.fill(7);
+        assert_eq!(vec2d[0], [7, 7, 7]);
+        assert_eq!(vec2d[1], [7, 7, 7]);
+    }
+
+    #[test]
+    fn vec2d_index() {
+        let vec2d = Vec2D {
+            data: vec![0, 1, 2, 3, 4, 5],
+            rows: 3,
```
gendx, line 68: There are only 2 rows in the corresponding data. This is an invalid `Vec2D` representation.

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
+use std::ops::{Index, IndexMut};
+
+/// A 2 dimensional matrix in row-major order backed by a contiguous `Vec`
+#[derive(Debug)]
+pub struct Vec2D<T> {
+    data: Vec<T>,
```
gendx, line 6: Given that the `cols` and `rows` are immutable, this doesn't need a growable vector (with a capacity), but only a boxed slice `Box<[T]>`.

src/macros.rs
```diff
@@ -39,3 +39,36 @@ macro_rules! lzma_debug {
 macro_rules! lzma_info {
     ($($arg:tt)+) => {};
 }
+
+// Macro to repeat elements of N times in an array without Clone
```
gendx, line 43: This mentions `Clone`, but `BitTree` is already `Clone`.
I'd rather not import a macro from a random StackOverflow question (what is the license?).

Alternatively, copy pasting the `BitTree::new` N times in the source code is fine for me.

src/decode/vec2d.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<T> IndexMut<usize> for Vec2D<T> {
+    fn index_mut(&mut self, idx: usize) -> &mut Self::Output {
+        &mut self.data[(idx * self.cols)..]
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
```
gendx, line 55: Can you add unit tests for:
- `init()`,
- `index_mut()`,
- some `should_panic` in case of indexing, e.g. `vec2d[0][3]` should manic on a 2x3 matrix.

src/decode/lzma.rs
```diff
@@ -1,6 +1,7 @@
 use crate::decode::lzbuffer::{LzBuffer, LzCircularBuffer};
 use crate::decode::rangecoder;
 use crate::decode::rangecoder::RangeDecoder;
```
gendx, line 3: It'd make more sense to import BitTree directly and merge these two lines together.

src/decode/lzma.rs
```diff
@@ -1,6 +1,7 @@
...
+use crate::decode::vec2d::Vec2D;
 use crate::decompress::Options;
 use crate::decompress::UnpackedSize;
```
gendx, line 6: Likewise, could you merge these two lines to keep a consistent import style?

gendx (changes requested):
Nice to see some performance improvements here! Ultimately, const generics (#19) would go further into this direction.

src/decode/lzma.rs
chyyran: Moved to `crate::util::vec2d`

src/macros.rs
chyyran: brainfart, I meant `Copy`, which `BitTree` is not. I've just copy-pasted `BitTree::new` N times instead if you're fine with that.

src/decode/vec2d.rs
chyyran: Since `rows` and `cols` are variable length and not `const`, there's no way to construct the array without const generics or [placement `in`](https://github.com/rust-lang/rust/issues/22181)

src/decode/vec2d.rs
chyyran: added unit tests for a variety of edge cases for `IndexMut` and `Index`

src/decode/vec2d.rs
chyyran: never mind, forgot about `into_boxed_slice`

chyyran:
I found that it was faster to just recreate stack-allocated arrays in `DecoderState` than bother with filling them. Heap-backed arrays still use `fill` to keep their allocation.

chyyran:
b90b484 removes the explicit bounds checks in `Index` and `IndexMut` for `Vec2D` as well as the `rows` argument which isn't necessary afterwards. We can rely on the bounds checks for `slice` that can be better optimized by rustc instead on Release.

Looks like clippy is failing on an unrelated file with a new lint, so I'm going to leave that as is.

src/decode/lzma.rs
```diff
@@ -202,33 +206,33 @@ impl DecoderState {
...
+        self.pos_decoders = [0x400; 115];
+        self.is_match = [0x400; 192];
+        self.is_rep = [0x400; 12];
+        self.is_rep_g0 = [0x400; 12];
+        self.is_rep_g1 = [0x400; 12];
+        self.is_rep_g2 = [0x400; 12];
+        self.is_rep_0long = [0x400; 192];
         self.state = 0;
-        self.rep.fill(0);
+        self.rep = [0; 4];
```
gendx, line 235: I don't think this brings any readability benefit over `fill`. The compiler should be able to optimize both versions of the code regardless.

I see you however observed otherwise:

> I found that it was faster to just recreate stack-allocated arrays in DecoderState than bother with filling them. Heap-backed arrays still use fill to keep their allocation.

Can you add a comment here to explain that?

(IMO, this might also be worth filing a bug to the Rust compiler, to understand if `fill` being slower is intentional or not a missed optimization somewhere in the compiler.)

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
+use std::ops::{Index, IndexMut};
+
+/// A 2 dimensional matrix in row-major order backed by a contiguous `Vec`
```
gendx, line 3: It's rather backed by a slice now.

Nit: also add a period `.` at the end of the sentence.

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+impl<T> Vec2D<T> {
+    /// Initialize a grid of size (`rows`, `cols`) with the given data element.
+    pub fn init(data: T, size: (usize, usize)) -> Vec2D<T>
+    where
+        T: Clone,
+    {
+        let (rows, cols) = size;
+        let len = rows
+            .checked_mul(cols)
+            .unwrap_or_else(|| panic!("{} rows by {} cols exceeds usize::MAX", rows, cols));
```
gendx, line 19: `unwrap_or` + `panic` can be replaced by `expect` instead (with `format!` here as the message isn't constant).

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+        vec2d.fill(7);
+        assert_eq!(vec2d[0], [7, 7, 7]);
+        assert_eq!(vec2d[1], [7, 7, 7]);
+    }
+
+    #[test]
+    fn vec2d_index() {
+        let vec2d = Vec2D {
+            data: vec![0, 1, 2, 3, 4, 5, 6, 7, 8].into_boxed_slice(),
+            cols: 3,
```
gendx, line 76: I wouldn't use the same number of rows and columns in this test, so that it's obvious in this test what are the rows vs. column dimensions.

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+        assert_eq!(vec2d[0], [0, 1, 2]);
+        assert_eq!(vec2d[1], [3, 9, 5]);
+        assert_eq!(vec2d[2], [6, 7, 8]);
+    }
+
+    #[test]
+    #[should_panic]
+    fn vec2d_index_out_of_bounds() {
+        let vec2d = Vec2D::init(1, (2, 3));
+        let _x = vec2d[2][4];
```
gendx, line 100: I'd use the minimal out-of-bounds index, i.e. `[2][3]`.

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+        &mut self.data[start_row..start_row + self.cols]
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn vec2d_init() {
```
gendx, line 58: The `vec2d_` prefix adds noise - all tests will appear nested within the `vec2d::test::` module anyway.

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[test]
+    #[should_panic]
+    fn vec2d_index_out_of_bounds_vec_edge() {
+        let vec2d = Vec2D::init(1, (2, 3));
+        let _x = vec2d[1][3];
+    }
+
+    #[test]
+    #[should_panic]
+    fn vec2d_index_out_of_bounds_overflow() {
```
gendx, line 112: I'd simply name this test `column_out_of_bounds`.

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+        let vec2d = Vec2D::init(1, (2, 3));
+        let _x = vec2d[0][3];
+    }
+
```
gendx, line 116: Can you also add a `row_out_of_bounds` test, i.e. `[2][0]`?

src/util/vec2d.rs
```diff
@@ -0,0 +1,137 @@
...
+impl<T> IndexMut<usize> for Vec2D<T> {
+    #[inline]
+    fn index_mut(&mut self, row: usize) -> &mut Self::Output {
+        let start_row = row * self.cols;
+        &mut self.data[start_row..start_row + self.cols]
+    }
+}
+
+#[cfg(test)]
+mod test {
```
gendx, line 54: Can you add a unit test for `init(0u8, (usize::MAX, usize::MAX))`, to test the multiplication overflow?

gendx (changes requested):
I've fixed the lints in #80, please rebase and re-run the workflows.

chyyran:
Addressed nits, and reformatted imports as per https://github.com/gendx/lzma-rs/pull/78#discussion_r953042701 as well.

src/util/vec2d.rs
chyyran: `unwrap_or_else` + `panic` is what clippy suggests to do to avoid `format!` being eagerly evaluated. There is unfortunatly not an `expect_or_else` function available.

```
error: use of `expect` followed by a function call
  --> src/util/vec2d.rs:[17](https://github.com/gendx/lzma-rs/runs/7986381449?check_suite_focus=true#step:4:18):42
   |
17 |           let len = rows.checked_mul(cols).expect(&format!(
   |  __________________________________________^
[18](https://github.com/gendx/lzma-rs/runs/7986381449?check_suite_focus=true#step:4:19) | |             "{} rows by {} cols exceeds usize::MAX",
[19](https://github.com/gendx/lzma-rs/runs/7986381449?check_suite_focus=true#step:4:20) | |             rows, cols
[20](https://github.com/gendx/lzma-rs/runs/7986381449?check_suite_focus=true#step:4:21) | |         ));
   | |__________^
   |
   = note: `-D clippy::expect-fun-call` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
help: try this
   |
17 ~         let len = rows.checked_mul(cols).unwrap_or_else(|| panic!("{} rows by {} cols exceeds usize::MAX",
18 ~             rows, cols));
   |
   ```

**Comment by chyyran** [comment]
Accidentally closed because of a faulty rebase, 444e68c106c65bde416d3f8baef4c184f7a613c7 should be properly rebased on current master now.

File: src/util/vec2d.rs [hunk 21]
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+impl<T> Index<usize> for Vec2D<T> {
+    type Output = [T];
+
+    #[inline]
+    fn index(&self, row: usize) -> &Self::Output {
+        let start_row = row * self.cols;
+        &self.data[start_row..start_row + self.cols]
```
gendx, line 41: Here we don't explicitly check that `start_row < rows`, so this addition may overflow, for example:

```rust
// Matrix with 5 columns.
let matrix = Vec2D::init(0, (3, 5));
// Somehow, as long as numbits(usize) is a multiple of 4, then 5 divides usize::MAX.
// This is clear in hexadecimal: usize::MAX is 0xFFF...F and usize::MAX / 5 is 0x333...3.
let row = usize::MAX / 5;
// This will therefore try to index data[usize::MAX..4].
matrix[row];
```

In practice this looks to be already caught by Rust's `std::slice::index`, but please add a `should_panic` test for this as well.

gendx (changes requested):
I've re-run the benchmarks, which show no significant change. However this is still a valuable cleanup - assuming the `Vec2D` implementation is correct with all the needed bounds checks!

src/util/vec2d.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.data.fill(value)
+    }
+}
+
+impl<T> Index<usize> for Vec2D<T> {
+    type Output = [T];
+
+    #[inline]
+    fn index(&self, row: usize) -> &Self::Output {
+        let start_row = row * self.cols;
```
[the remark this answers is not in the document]
chyyran: Fixed in 48b66fa8b82335f17d5d4ec5a7e2531114f40f2c

src/util/vec2d.rs
```diff
@@ -0,0 +1,160 @@
...
+    #[test]
+    #[should_panic]
+    fn mut_column_out_of_bounds() {
+        let mut vec2d = Vec2D::init(1, (2, 3));
+        vec2d[0][3] = 0;
+    }
+
+    #[test]
+    #[should_panic]
+    fn indexmut_out_of_bounds() {
```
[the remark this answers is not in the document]
chyyran: Fixed in 48b66fa8b82335f17d5d4ec5a7e2531114f40f2c

src/util/vec2d.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn index(&self, row: usize) -> &Self::Output {
+        let start_row = row * self.cols;
+        &self.data[start_row..start_row + self.cols]
+    }
+}
+
+impl<T> IndexMut<usize> for Vec2D<T> {
+    #[inline]
+    fn index_mut(&mut self, row: usize) -> &mut Self::Output {
+        let start_row = row * self.cols;
```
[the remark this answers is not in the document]
chyyran: Fixed in 48b66fa8b82335f17d5d4ec5a7e2531114f40f2c

src/util/vec2d.rs
chyyran: Added a test for this which panics correctly, but did not bother with `checked_add`.

chyyran:
Fixed nits in 48b66fa8b82335f17d5d4ec5a7e2531114f40f2c

gendx (approved):

PR #78, opened 2022-08-09
Update dependencies (particularly crc 3.0.0)

chyyran:
### Pull Request Overview

This pull request adds/changes/fixes...

- Updates `byteorder` to `1.4.3`
- Updates `log` to `0.4.17`
- Updates `env_logger` to `0.9.0`
- Updates `crc` to `3.0.0`
  - `crc` from `^1.0.0` -> `3.0.0` represents a very large API change. In particular, algorithms are now const-initialized and `Digest` no longer implements `Hash`, so `Write/Read` for `DigestWrite/Read` (formerly `HasherWrite/Read`) has to be implemented per type. This results in most of the code changes in this PR. 
  - The particular selection of CRC polynomials are consistent with the old version of `crc` and the XZ specification. See [CRC-64/XZ](https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-64-xz) and and [CRC-32/ISO-HDLC](https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-32-iso-hdlc) in the catalogue of parameterised CRC algorithms.

### Testing Strategy

This pull request was tested by ...
- [x] The existing tests are sufficient and should all pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,10 @@ exclude = ["tests/*", "benches/*", "fuzz/*", ".github/*", "Cargo.lock"]
 edition = "2018"
 
 [dependencies]
-byteorder = "^1.0.0"
-crc = "^1.0.0"
-log = { version = "^0.4.14", optional = true }
-env_logger = { version = "^0.8.3", optional = true }
+byteorder = "1.4.3"
+crc = "3.0.0"
+log = { version = "0.4.17", optional = true }
+env_logger = { version = "0.9.0", optional = true }
 
 [dev-dependencies]
 rust-lzma = "0.5"
--- a/src/decode/util.rs
+++ b/src/decode/util.rs
@@ -1,4 +1,4 @@
-use std::{hash, io};
+use std::io;
 
 pub fn read_tag<R: io::BufRead>(input: &mut R, tag: &[u8]) -> io::Result<bool> {
     let mut buf = vec![0; tag.len()];
@@ -34,33 +34,32 @@ pub fn flush_zero_padding<R: io::BufRead>(input: &mut R) -> io::Result<bool> {
 }
 
 // A Read computing a digest on the bytes read.
-pub struct HasherRead<'a, R, H>
+pub struct CrcDigestRead<'a, 'b, R, S>
 where
     R: 'a + io::Read,
-    H: 'a + hash::Hasher,
+    S: crc::Width,
 {
-    read: &'a mut R,   // underlying reader
-    hasher: &'a mut H, // hasher
+    read: &'a mut R,                    // underlying reader
+    digest: &'a mut crc::Digest<'b, S>, // hasher
 }
 
-impl<'a, R, H> HasherRead<'a, R, H>
+impl<'a, 'b, R, S> CrcDigestRead<'a, 'b, R, S>
 where
     R: io::Read,
-    H: hash::Hasher,
+    S: crc::Width,
 {
-    pub fn new(read: &'a mut R, hasher: &'a mut H) -> Self {
-        Self { read, hasher }
+    pub fn new(read: &'a mut R, digest: &'a mut crc::Digest<'b, S>) -> Self {
+        Self { read, digest }
     }
 }
 
-impl<'a, R, H> io::Read for HasherRead<'a, R, H>
+impl<'a, 'b, R> io::Read for CrcDigestRead<'a, 'b, R, u32>
 where
     R: io::Read,
-    H: hash::Hasher,
 {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         let result = self.read.read(buf)?;
-        self.hasher.write(&buf[..result]);
+        self.digest.update(&buf[..result]);
         Ok(result)
     }
 }
--- a/src/decode/xz.rs
+++ b/src/decode/xz.rs
@@ -3,10 +3,9 @@
 use crate::decode::lzma2::Lzma2Decoder;
 use crate::decode::util;
 use crate::error;
+use crate::xz::crc::{CRC32, CRC64};
 use crate::xz::{footer, header, CheckMethod, StreamFlags};
 use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
-use crc::{crc32, crc64, Hasher32};
-use std::hash::Hasher;
 use std::io;
 use std::io::Read;
 
@@ -46,11 +45,9 @@ where
     };
 
     let crc32 = input.read_u32::<LittleEndian>()?;
-    let mut digest = crc32::Digest::new(crc32::IEEE);
-
+    let mut digest = CRC32.digest();
     {
-        let mut digested = util::HasherRead::new(input, &mut digest);
-
+        let mut digested = util::CrcDigestRead::new(input, &mut digest);
         let backward_size = digested.read_u32::<LittleEndian>()?;
         if index_size as u32 != (backward_size + 1) << 2 {
             return Err(error::Error::XzError(format!(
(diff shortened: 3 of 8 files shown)


src/decode/util.rs
```diff
@@ -35,33 +34,32 @@ pub fn flush_zero_padding<R: io::BufRead>(input: &mut R) -> io::Result<bool> {
...
 {
-    pub fn new(read: &'a mut R, hasher: &'a mut H) -> Self {
-        Self { read, hasher }
+    pub fn new(read: &'a mut R, digest: &'a mut crc::Digest<'b, S>) -> Self {
+        Self { read, digest }
     }
 }
 
-impl<'a, R, H> io::Read for HasherRead<'a, R, H>
+impl<'a, 'b, R> io::Read for DigestRead<'a, 'b, R, u32>
```
chyyran, line 56: Because of how [`impl Digest`](https://docs.rs/crc/latest/crc/struct.Digest.html) is defined there's no way to define it once for all possible types.

In practice we only need `u32` for `impl Read` for usage in lzma-rs, so it's only implemented for `u32` here.

src/encode/util.rs
```diff
@@ -1,37 +1,34 @@
...
 {
-    pub fn new(write: &'a mut W, hasher: &'a mut H) -> Self {
-        Self { write, hasher }
+    pub fn new(write: &'a mut W, digest: &'a mut crc::Digest<'b, S>) -> Self {
+        Self { write, digest }
     }
 }
 
-impl<'a, W, H> io::Write for HasherWrite<'a, W, H>
+impl<'a, 'b, W> io::Write for DigestWrite<'a, 'b, W, u32>
```
chyyran, line 23: Because of how [`impl Digest`](https://docs.rs/crc/latest/crc/struct.Digest.html) is defined there's no way to define it once for all possible types.

In practice we only need `u32` for `impl Write` for usage in lzma-rs, so it's only implemented for `u32` here.

src/decode/util.rs
```diff
@@ -35,33 +34,32 @@ pub fn flush_zero_padding<R: io::BufRead>(input: &mut R) -> io::Result<bool> {
 }
 
 // A Read computing a digest on the bytes read.
-pub struct HasherRead<'a, R, H>
+pub struct DigestRead<'a, 'b, R, S>
```
chyyran, line 37: `S` for "Size", to be consistent with `DigestWrite`.

chyyran, line 37: Renamed to `DigestRead` because `Digest` no longer implements `Hash` so we can not provide a generic implementation.

src/encode/util.rs
```diff
@@ -1,37 +1,34 @@
-use std::hash;
 use std::io;
 
 // A Write computing a digest on the bytes written.
-pub struct HasherWrite<'a, W, H>
+pub struct DigestWrite<'a, 'b, W, S>
```
chyyran, line 4: Renamed to `DigestWrite` because `Digest` no longer implements `Hash` so we can not provide a generic implementation.

chyyran, line 4: `S` for "Size", since `Write` already took `W`

src/decode/util.rs
chyyran, line 37: Extra `'b` lifetime is specifies that the lifetime of the mutable borrows (`'a`) is different from the lifetime of the CRC algorithm (`'b`) which `Digest` borrows immutably.

src/encode/util.rs
chyyran, line 4: Extra `'b` lifetime is specifies that the lifetime of the mutable borrows (`'a`) is different from the lifetime of the CRC algorithm (`'b`) which `Digest` borrows immutably.

Cargo.toml
```diff
@@ -12,10 +12,10 @@ exclude = ["tests/*", "benches/*", "fuzz/*", ".github/*", "Cargo.lock"]
 edition = "2018"
 
 [dependencies]
-byteorder = "^1.0.0"
-crc = "^1.0.0"
-log = { version = "^0.4.14", optional = true }
-env_logger = { version = "^0.8.3", optional = true }
+byteorder = "^1.4.3"
+crc = "3.0.0"
```
gendx, line 16: Let's use a consistent style: either a caret everywhere or nowhere.

src/decode/util.rs
gendx: In the end, due to this limitation, the old `HasherRead` is not as generic anymore, and only works for `digest == CRC32`. I see two ways to resolve this design conflict:

1. Keep most of the original implementation, but replace `hash::Hasher` by a new `Digest` trait that only contains `fn update(&mut self, bytes: &[u8])` (without the `finalize` part) - and add an `impl Digest for crc::Digest<_>`. That way it's still compatible with all the CRC sizes, as well other hash algorithms.
2. Remove the genericity over the digest type, and rebrand the `HasherRead` as `Crc32DigestRead` (with only genericity over the `io::Read`).

I'd lean towards solution (1) to minimize code diff and to unlock potential support for other digest algorithms, but (2) would be fine as well.

src/decode/xz.rs
```diff
@@ -3,10 +3,11 @@
 use crate::decode::lzma2::Lzma2Decoder;
 use crate::decode::util;
 use crate::error;
-use crate::xz::{footer, header, CheckMethod, StreamFlags};
+use crate::xz::{
+    crc::{CRC32, CRC64},
+    footer, header, CheckMethod, StreamFlags,
+};
```
gendx, line 9: The implicit style so far was to format imports with `imports_granularity="Module"` (https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#module). Could you format the new imports accordingly?

```rust
use crate::xz::crc::{CRC32, CRC64};
use crate::xz::{footer, header, CheckMethod, StreamFlags};
```

I'll separately look into how to update the format lint check to use this (unstable) rustfmt option.

gendx (changes requested):
Overall a good change, with some minor comments.

src/decode/util.rs
chyyran: `DigestRead`/`DigestWrite` is still generic over other CRC sizes, there just needs to be an impl for sizes other than `u32` which I've omitted for YAGNI reasons; adding sizes other than CRC32 is trivial where the implementation is exactly the same. If you prefer, we can impl it for all known CRC sizes with a macro.  What it is no longer generic over is however, all implementations of `Hasher` that are not CRC. Again, for YAGNI reasons I don't think there is currently a strong case for (1).

Instead, I've went with (2) and clarified the naming as `CrcDigestRead`/`CrcDigestWrite`.

Cargo.toml
chyyran: Removed carets from Cargo.toml

chyyran:
Addressed the nits and rebased on master. See follow up comment on reasoning for `CrcDigestRead/Write`.

gendx (changes requested):

src/xz/header.rs
```diff
@@ -28,13 +28,13 @@ impl StreamHeader {
             )));
         }
 
-        let (flags, digested) = {
-            let mut digest = crc32::Digest::new(crc32::IEEE);
-            let mut digest_rd = util::HasherRead::new(input, &mut digest);
-            let value = digest_rd.read_u16::<BigEndian>()?;
-            (value, digest.sum32())
+        let mut digest = CRC32.digest();
```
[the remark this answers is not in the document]
chyyran: Fixed in 4d800af

gendx:
In the meantime I've merged #82 which addresses the import formatting question.

Could you rebase on top of it to resolve the merge conflict?

chyyran:
Rebased on master in 698620b

PR #79, opened 2022-08-09
Use const generics to remove BitTree heap allocations

### LINKED ISSUE #19: Use const-generics

chyyran:
### Pull Request Overview

#19 was blocked on `#[feature(generic_const_exprs)]` for the `1 << NUM_BITS` issue but we do not actually need `generic_const_exprs` to utilize const generics at the expense of having a slightly surprising API for `BitTree`. Since `BitTree` is an internal implementation detail, I do not think it is too much of a trade-off for the performance benefits seen. The strong compile time guarantee of the validity of `BitTree` when using `NUM_BITS` in an expression can be replicated with a compile time assert in the constructor.

~~There are **2 different API shapes** for `BitTree` proposed here. Both APIs validate their const arguments at compile time, and invalid instances of `BitTree` are still unconstructable.~~

~~1. `BitTree<const NUM_BITS: usize, const PROBS_ARRAY_LEN: usize>` (06aee699384477e09aa98e3d029197e927536b31)~~
2. `BitTree<const PROBS_ARRAY_LEN: usize>` (260eefe24b47e07d8cff2ccc8afda36755ce5ca2)

The first signature must have the caller input both `NUM_BITS`, and `PROBS_ARRAY_LEN == 1 << NUM_BITS` as part of the type signature. The equality of `PROBS_ARRAY_LEN` is checked by a compile-time assert macro that simply checks that indeed `PROBS_ARRAY_LEN == 1 << NUM_BITS`. This prevents the construction of a `BitTree` where `PROBS_ARRAY_LEN` is not `1 << NUM_BITS`

The second signature takes only `PROBS_ARRAY_LEN` to sidestep the requirement for `generic_const_exprs`. Instead, `NUM_BITS` is an associated constant that is calculated as `floor(log_2(PROBS_ARRAY_LEN))` using `PROBS_ARRAY_LEN.trailing_zeros()` which has been `const` since Rust 1.32. The const assert instead checks that `PROBS_ARRAY_LEN == 1 << (PROBS_ARRAY_LEN.trailing_zeros())`. Thus, rather than `BitTree` being valid for any `NUM_BITS` and `PROBS_ARRAY_LEN == 1 << NUM_BITS`, `BitTree` in this API is valid for any `PROBS_ARRAY_LEN` such that `PROBS_ARRAY_LEN == 2 ** floor(log_2(PROBS_ARRAY_LEN))`, which gives us effectively the same guarantees as the first signature without the redundant `NUM_BITS` argument. 

Both signatures compile to the same code and have the same performance. It is a matter of preference and readability on which to take. I have a slight preference for the second signature because it is less noisy and seeing `BitTree<{1 << 8}>` for example seems fairly obvious, but purely looking at the implementation, I prefer the less 'clever' first signature.

This PR bumps the MSRV to 1.57 for const generics and const_panic.

#### Benchmarks

In general, variances are reduced and there is a speed bump when accessing the dictionary during decompression. While I did apply const-generics to `encode::rangecoder::BitTree` for consistency, they don't seem to be currently used for compression so compression benchmarks should not be heavily affected if at all. 

Without const generics
```
running 8 tests
test compress_65536                  ... bench:   1,312,743 ns/iter (+/- 45,836)
test compress_empty                  ... bench:         712 ns/iter (+/- 55)
test compress_hello                  ... bench:       1,022 ns/iter (+/- 18)
test decompress_after_compress_65536 ... bench:   1,157,757 ns/iter (+/- 64,077)
test decompress_after_compress_empty ... bench:       1,897 ns/iter (+/- 82)
test decompress_after_compress_hello ... bench:       2,256 ns/iter (+/- 165)
test decompress_big_file             ... bench:   3,615,488 ns/iter (+/- 128,563)
test decompress_huge_dict            ... bench:       2,318 ns/iter (+/- 135)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out; finished in 16.21s
```

With const generic BitTree
```
test compress_65536                  ... bench:   1,317,185 ns/iter (+/- 58,197)
test compress_empty                  ... bench:         791 ns/iter (+/- 65)
test compress_hello                  ... bench:       1,108 ns/iter (+/- 45)
test decompress_after_compress_65536 ... bench:   1,189,749 ns/iter (+/- 60,014)
test decompress_after_compress_empty ... bench:         530 ns/iter (+/- 23)
test decompress_after_compress_hello ... bench:         787 ns/iter (+/- 26)
test decompress_big_file             ... bench:   3,600,336 ns/iter (+/- 375,302)
test decompress_huge_dict            ... bench:         873 ns/iter (+/- 33)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out; finished in 11.17s
```

This should fix #19, after this PR, `DecoderState` only has the single heap allocation of `literal_probs` which is substantially more difficult to put on the stack. 

### Testing Strategy
* This is a backend change and does not introduce additional functionality, existing tests should pass.

### DIFF
--- a/src/decode/rangecoder.rs
+++ b/src/decode/rangecoder.rs
@@ -1,5 +1,6 @@
 use crate::decode::util;
 use crate::error;
+use crate::util::const_assert;
 use byteorder::{BigEndian, ReadBytesExt};
 use std::io;
 
@@ -150,27 +151,42 @@ where
     }
 }
 
-// TODO: parametrize by constant and use [u16; 1 << num_bits] as soon as Rust supports this
 #[derive(Debug, Clone)]
-pub struct BitTree {
-    num_bits: usize,
-    probs: Vec<u16>,
+pub struct BitTree<const PROBS_ARRAY_LEN: usize> {
+    probs: [u16; PROBS_ARRAY_LEN],
 }
 
-impl BitTree {
-    pub fn new(num_bits: usize) -> Self {
+impl<const PROBS_ARRAY_LEN: usize> BitTree<PROBS_ARRAY_LEN> {
+    pub fn new() -> Self {
+        // The validity of PROBS_ARRAY_LEN is checked at compile-time with a macro
+        // that confirms that the argument P passed is indeed 1 << N for
+        // some N using usize::trailing_zeros to calculate floor(log_2(P)).
+        //
+        // Thus, BitTree<const P: usize> is only valid for any P such that
+        // P = 2 ** floor(log_2(P)), where P is the length of the probability array
+        // of the BitTree. This maintains the invariant that P = 1 << N.
+        //
+        // This precondition must be checked for any way to construct a new, valid instance of BitTree.
+        // Here it is checked for BitTree::new(), but if another function is added that returns a
+        // new instance of BitTree, this assertion must be checked there as well.
+        const_assert!("BitTree's PROBS_ARRAY_LEN parameter must be a power of 2",
+            PROBS_ARRAY_LEN: usize => (1 << (PROBS_ARRAY_LEN.trailing_zeros() as usize)) == PROBS_ARRAY_LEN);
         BitTree {
-            num_bits,
-            probs: vec![0x400; 1 << num_bits],
+            probs: [0x400; PROBS_ARRAY_LEN],
         }
     }
+
+    // NUM_BITS is derived from PROBS_ARRAY_LEN because of the lack of
+    // generic const expressions. Where PROBS_ARRAY_LEN is a power of 2,
+    // NUM_BITS can be derived by the number of trailing zeroes.
+    const NUM_BITS: usize = PROBS_ARRAY_LEN.trailing_zeros() as usize;
 
     pub fn parse<R: io::BufRead>(
         &mut self,
         rangecoder: &mut RangeDecoder<R>,
         update: bool,
     ) -> io::Result<u32> {
-        rangecoder.parse_bit_tree(self.num_bits, self.probs.as_mut_slice(), update)
+        rangecoder.parse_bit_tree(Self::NUM_BITS, &mut self.probs, update)
     }
 
     pub fn parse_reverse<R: io::BufRead>(
@@ -178,11 +194,7 @@ impl BitTree {
         rangecoder: &mut RangeDecoder<R>,
         update: bool,
     ) -> io::Result<u32> {
-        rangecoder.parse_reverse_bit_tree(self.num_bits, self.probs.as_mut_slice(), 0, update)
-    }
-
-    pub fn reset(&mut self) {
-        self.probs.fill(0x400);
+        rangecoder.parse_reverse_bit_tree(Self::NUM_BITS, &mut self.probs, 0, update)
     }
 }
 
@@ -190,9 +202,9 @@ impl BitTree {
 pub struct LenDecoder {
     choice: u16,
     choice2: u16,
-    low_coder: [BitTree; 16],
-    mid_coder: [BitTree; 16],
-    high_coder: BitTree,
+    low_coder: [BitTree<{ 1 << 3 }>; 16],
+    mid_coder: [BitTree<{ 1 << 3 }>; 16],
+    high_coder: BitTree<{ 1 << 8 }>,
 }
 
 impl LenDecoder {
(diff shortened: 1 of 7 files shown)


chyyran:
https://github.com/gendx/lzma-rs/pull/79/commits/37e09d126b83bbb390763af5a24626c3ab1dde98 represents an alternative API with `BitTree` parameterized on a single generic, which is unfortunately not `NUM_BITS` but `PROBS_ARRAY_LEN`  as we can't use associated constants in struct definitions. This relies on the idea that that `1 << n` has the inverse `(1 << n).trailing_zeroes()`, both of which are const. We can then use `NUM_BITS` as an associated constant.

The signature of `BitTree` here is cleaner, but the downside is that it is now possible to construct invalid instances of `BitTree` since there is no way to check that `PROBS_ARRAY_LEN` is the intended value without passing in `NUM_BITS`, and there is no way to check for this. For this reason I prefer the former implementation, but since `BitTree` is an internal API this really is just a matter of style and clarity.

chyyran:
https://github.com/gendx/lzma-rs/pull/79/commits/9223b59a1fd13b28bb88aa27a632f7dbc867fd9c introduces a combining of both approaches. While the generic argument is still `1 << NUM_BITS` (removed the helper const function since I thought it was clear enough without it), I realized there can still be some safety had to ensure that as long as `trailing_zeros` was valid for log_2 the input array length would be valid.

src/encode/rangecoder.rs
```diff
@@ -274,40 +319,43 @@ mod test {
 
     #[test]
     fn test_encode_decode_bittree_zeros() {
-        for num_bits in 0..16 {
-            encode_decode_bittree(num_bits, &[0; 10000]);
-        }
+        seq!(NUM_BITS in 0..16 {
```
chyyran, line 322: Since `NUM_BITS` is now compile-time, the `seq!` macro is used for these tests instead.

chyyran:
Marking this as ready since #77 was merged.

@gendx you will want to review both 06aee699384477e09aa98e3d029197e927536b31 and 260eefe24b47e07d8cff2ccc8afda36755ce5ca2 which are identical **except for the signature of `BitTree`**, as explained in the overview. 

Given that `generic_const_exprs` doesn't seem to be coming very soon, and that `BitTree` is an internal API that isn't exposed, I'm of the opinion that the perf improvements are worth the hit to readability, but you may want to consider re-running some benchmarks on your end as well.

src/util/assert.rs
```diff
@@ -0,0 +1,15 @@
...
+    ($($list:ident : $ty:ty),* => $expr:expr) => {{
+        struct Assert<$(const $list: $ty,)*>;
+        impl<$(const $list: $ty,)*> Assert<$($list,)*> {
+            const OK: u8 = 0 - !($expr) as u8;
+        }
+        Assert::<$($list,)*>::OK
+    }};
+    ($expr:expr) => {
+        const OK: u8 = 0 - !($expr) as u8;
+    };
```
gendx, line 12: Is this branch ever needed?

src/decode/rangecoder.rs
```diff
@@ -150,49 +151,52 @@ where
...
+impl<const PROBS_ARRAY_LEN: usize> BitTree<PROBS_ARRAY_LEN> {
+    pub fn new() -> Self {
+        // The validity of PROBS_ARRAY_LEN is checked at compile-time with a macro
+        // that confirms that the argument P passed is indeed 1 << N for
+        // some N using usize::trailing_zeros to calculate floor(log_2(P)).
+        //
+        // Thus, BitTree<const P: usize> is only valid for any P such that
+        // P = 2 ** floor(log_2(P)), where P is the length of the probability array
+        // of the BitTree. This maintains the invariant that P = 1 << N.
+        const_assert!(PROBS_ARRAY_LEN: usize => (1 << (PROBS_ARRAY_LEN.trailing_zeros() as usize)) == PROBS_ARRAY_LEN);
```
gendx, line 168: One problem with this approach is that the assertion won't trigger if the `new()` function is not called on a given parameter: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f245e4d1da737e4e9484d96a5187b988

I guess it's fine in practice, but could you explain that in a comment?

gendx (changes requested):
Apologies for the delay in the last few months.

Overall this PR looks good, I have some comments mostly regarding the static assertion macro.

Apart from that, I've just added code coverage reporting (https://github.com/gendx/lzma-rs/pull/86), so let's see how this PR works with it.

src/util/assert.rs
chyyran: Removed

src/util/assert.rs
```diff
@@ -0,0 +1,15 @@
+/// macro for compile-time const assertions
+macro_rules! const_assert {
+    ($($list:ident : $ty:ty),* => $expr:expr) => {{
+        struct Assert<$(const $list: $ty,)*>;
+        impl<$(const $list: $ty,)*> Assert<$($list,)*> {
+            const OK: u8 = 0 - !($expr) as u8;
```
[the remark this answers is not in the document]
chyyran: Addressed in [025c331](https://github.com/gendx/lzma-rs/pull/79/commits/025c33142e265080bafc951e256f7a0fb5ede3b6) but this will bump MSRV from 1.51 to 1.57.

chyyran:
I've rebased on master, fixing the conflicts and squashed the commits into one. Looks like there's a small improvement in code coverage as well!

chyyran:
@gendx Any updates on getting this merged?

src/util/assert.rs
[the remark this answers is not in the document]
gendx: Bumping to 1.57 should be fine by now. I've also cut a new version 0.3.0 (https://github.com/gendx/lzma-rs/commit/da82bd1182993a07db035e285a8f8a87d32a30c2), so the 0.2.x branch is still available for compilers < 1.57.

gendx:
Apologies again for the delay. I'll merge this now and fix any pending issue on top of it.

Thanks a lot for your contribution!

gendx (approved):

PR #90, opened 2022-11-20
allow user to ignore trailing bytes

Niedzwiedzw:
### Pull Request Overview
it allows for user to ignore trailing bytes, which made the crate unusable for me personally

### Testing Strategy

### DIFF
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -165,6 +165,7 @@ pub(crate) struct DecoderState {
     // MAX_REQUIRED_INPUT bytes can be consumed during one iteration.
     partial_input_buf: std::io::Cursor<[u8; MAX_REQUIRED_INPUT]>,
     pub(crate) lzma_props: LzmaProperties,
+    pub(crate) decoder_options: Options,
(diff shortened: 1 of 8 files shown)


Niedzwiedzw:
I have also considered feature flag given that this changes the API a bit, I'll leave it up to your judgement

7ERr0r:
That's a breaking change. Your API functions have different params.
I would suggest:
- renaming your `LzmaDecoder::new` to maybe `LzmaDecoder::new_with_options`
- renaming `allow_trailing_bytes` to `allow_trailing_bytes_lzma` since it's not effective for LZMA2 and XZ
- maybe separating `Options` into `LzmaOptions` from `StreamOptions`

gendx:
- Breaking API changes are not a problem per se, as they can be guarded by features and a SemVer bump at the next release.
- As far as I understand it, this new option only applies to the raw stream, and therefore doesn't need to be part of the LZMA2 and xz APIs (which can simply invoke the LZMA layer with the default option).
- In any case, unit tests are needed.

PR #95, opened 2023-01-28
lib: enable no-std compatibility

rmsyn:
### Pull Request Overview

Add a default `std` feature, and use compatible data structures from `alloc`, `core`, and `core2` for `no-std` builds.

To build for `no-std` pass `--no-default-features` on the command line, or `default-features = false` in the Cargo.toml of another library/binary using `lzma-rs`.

### Testing Strategy

Building and testing with the new `std` feature, and with `no-default-features`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ byteorder = "1.4.3"
 crc = "3.0.0"
 log = { version = "0.4.17", optional = true }
 env_logger = { version = "0.9.0", optional = true }
+core2 = { version = "^0.4", optional = true }
 
 [dev-dependencies]
 rust-lzma = "0.5"
@@ -25,6 +26,7 @@ seq-macro = "0.3"
 enable_logging = ["env_logger", "log"]
 stream = []
 raw_decoder = []
+no_std = ["core2"]
 
 [package.metadata.docs.rs]
 features = ["stream", "raw_decoder"]
--- a/src/decode/lzbuffer.rs
+++ b/src/decode/lzbuffer.rs
@@ -1,4 +1,7 @@
 use crate::error;
+#[cfg(feature = "no_std")]
+use core2::io;
+#[cfg(not(feature = "no_std"))]
 use std::io;
 
 pub trait LzBuffer<W>
--- a/src/decode/lzma.rs
+++ b/src/decode/lzma.rs
@@ -3,7 +3,12 @@ use crate::decode::rangecoder::{BitTree, LenDecoder, RangeDecoder};
 use crate::decompress::{Options, UnpackedSize};
 use crate::error;
 use crate::util::vec2d::Vec2D;
+#[cfg(feature = "no_std")]
+use alloc::string::String;
 use byteorder::{LittleEndian, ReadBytesExt};
+#[cfg(feature = "no_std")]
+use core2::io;
+#[cfg(not(feature = "no_std"))]
 use std::io;
 
 /// Maximum input data that can be processed in one iteration.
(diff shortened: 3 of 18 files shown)


Cargo.toml
```diff
@@ -16,14 +16,17 @@ byteorder = "1.4.3"
 crc = "3.0.0"
 log = { version = "0.4.17", optional = true }
 env_logger = { version = "0.9.0", optional = true }
+core2 = "0.4"
```
gendx, line 19: This crate has not been updated in 2 years and has a "CI failing" badge, so I'm not sure it's a good choice.

Cargo.toml
```diff
@@ -16,14 +16,17 @@ byteorder = "1.4.3"
...
 
 [dev-dependencies]
 rust-lzma = "0.5"
 
 [features]
+default = ["std"]
 enable_logging = ["env_logger", "log"]
 stream = []
 raw_decoder = []
+std = []
```
gendx, line 29: In terms of features, "no_std" should be an opt in, as it pulls an additional crate (core2). This crate should be optional, and only pulled for users who explicitly opt into the "no_std" feature.

gendx (changes requested):

Cargo.toml
rmsyn: I can change this. Makes sense to do opt-in for pulling in optional dependency  crates.

Cargo.toml
rmsyn: I've opened a pull request to fix CI builds in [core2/#21](https://github.com/technocreatives/core2/pull/21).

Cargo.toml
```diff
@@ -25,6 +26,7 @@ seq-macro = "0.3"
 enable_logging = ["env_logger", "log"]
 stream = []
 raw_decoder = []
+no_std = ["core2"]
```
moridinga, line 29: This is 100% opposite of what is suggested in the [cargo book](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification):

"features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

For example, if you want to optionally support [no_std](https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute) environments, do not use a no_std feature. Instead, use a std feature that enables std."

In my experience, people are definitely following the cargo book's recommendation. So I think you had it right the first time.

rmsyn: Thanks for the review.

> Instead, use a std feature that enables std

I think this advice is for the general case, when optional dependencies aren't involved. It's not possible to use optional dependencies in a `no_std` build using the `std` feature approach.

This is one of the main reasons @gendx suggested using the `no_std` feature approach.

Reverting to the original PR isn't much work, but the changes do have their drawbacks.

> A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

This PR (both versions) do not break SemVer, users would still get `std` builds by default. Compiled code should literally be identical to code without the changes, unless the opt-in feature is used.

In the original, the user opts-in with `--default-features false`. In the current version, the user opts-in with `--features no_std`. Otherwise, the build should use all the `std` dependencies, and be identical to the current mainline.

moridinga: `std` should be part of your defaults, people who want to use it with no_std will use `--default-features false` (or ` default-features = false` in the Cargo.toml). This is the pattern that exists across all the crates I work with in the embedded space.

rmsyn: > `std` should be part of your defaults, people who want to use it with no_std will use `--default-features false` (or ` default-features = false` in the Cargo.toml). This is the pattern that exists across all the crates I work with in the embedded space.

I agree with you, that's why I did it that way in the original PR. The (now-)optional dependency gets dropped from compilation when `std` feature is enabled, since it's not used anywhere else. That essentially turns a required dependency into an optional dependency, with the exception that the crate (and its dependencies) are always downloaded.

I'm fine with either version, honestly. I can revert to the original changes, or keep the PR as it is now. It's up to what @gendx wants to do.

Still no movement on the `core2` PR.

PR #100, opened 2023-05-20
Small fix to RangeEncoder

WinLinux1028:
The rationale for this modification can be found below.
https://github.com/tukaani-project/xz-java/blob/master/src/org/tukaani/xz/rangecoder/RangeEncoder.java#L29
https://github.com/tukaani-project/xz-java/blob/master/src/org/tukaani/xz/rangecoder/RangeEncoder.java#L115

### DIFF
--- a/src/encode/rangecoder.rs
+++ b/src/encode/rangecoder.rs
@@ -12,7 +12,7 @@ where
     range: u32,
     low: u64,
     cache: u8,
-    cachesz: u32,
+    cachesz: u64,
 }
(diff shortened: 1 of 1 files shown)


gendx:
Thank you for the contribution.

- As far as I understand it, the main difference is about making the cache size a `u64`. The `normalize()` function doesn't need to be remove for this change.
- It would be good to have a test case that triggers the behavior, to avoid future regressions.

PR #116, opened 2025-07-16
Add support for `no_std`

kouchekiniad:
### Pull Request Overview

This change adds support for `no_std` environments that still support the `alloc` crate. To achieve this, all references to `std::io` have been directed to an internal `io` module. In the default case (with the "std" feature flag enabled) the new `io` module simply re-exports relevant `std::io` constructs, leaving all current functionality intact. With the "std" feature flag disabled, `io` implements a minimal set of `std::io`-like constructs that are used as drop-in replacements for the use of `std::io` features in the rest of the crate.

The `no_std` `io` implementation allows any `AsRef<u8>` (including `&[u8]` and `Vec<u8>`) to be used as a `BufRead` by wrapping it in a `io::Cursor` with `io::Cursor::new()`, and `Vec<u8>`, `io::Cursor<Vec<u8>>`, and `io::Cursor<&mut [u8]>` to be used as `Write`.

### Testing Strategy

- [x] Added relevant unit tests to the `no_std` `io` implementation in `nostd.rs`.
- [x] Modified workflow to run all tests and benchmarks with and without `std` support

Updated Rustdoc and README.md

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -8,5 +8,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@nightly
+      - name: Install liblzma-dev
+        if: runner.os == 'Linux'
+        run: sudo apt-get update && sudo apt-get install -y liblzma-dev
       - name: Build benches
-        run: cargo build --benches --verbose
+        run: cargo +nightly build --benches --verbose
+      - name: Build benches (no default features)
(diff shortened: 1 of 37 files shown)


kouchekiniad:
Hi @gendx, would you be willing to take this change? The change is effectively non-breaking, with the only visible API difference in `std` environments being that `io::BufRead` and `io::Write` are now referenced as being re-exported through the `io` module.

Issue #110, opened 2024-05-05
add `xz_decompress_with_options` API


jtmoon79:
# Problem

`xz_decompress` reads the entire file in one call, and thus the entire file is stored in memory.  Decompressing large files can lead to _Out Of Memory_ errors.

# Solution

Provide `xz_decompress_with_options`, similar to other `_with_options` API functions. Notably, this would allow passing an [Options](https://docs.rs/lzma-rs/latest/lzma_rs/decompress/struct.Options.html) with a set `memlimit`. Then the user need not store the entire file in memory.

PR #118, opened 2026-04-01
Feat/implement streaming decompression

Hahihula:
Implement true O(1) memory streaming for XZ decompression

### Summary
This PR fundamentally changes how `lzma-rs` handles XZ decompression under the hood, transitioning from buffering entire compressed blocks in memory to true constant-memory (O(1)) streaming. 

Previously, decompressing a large XZ file (e.g., 1GB) would allocate the entire uncompressed block into a `Vec<u8>` before writing it to the output. With this PR, memory usage is strictly bounded by the LZMA2 dictionary size (e.g., 64MB), regardless of the total file size.

### Motivation
The standard `xz` format allows for massive block sizes. Because `lzma-rs` was accumulating the entire block in an `LzAccumBuffer` before flushing it to the output writer, memory usage would spike wildly. This made the crate unsuitable for memory-constrained environments or decompressing very large files.

### Changes Made

**1. XZ Layer: Streaming Checksum Validation (`src/decode/xz.rs`)**
* Introduced `CrcWriter`, a wrapper around the output writer that calculates CRC32/CRC64 hashes on-the-fly as data is written.
* Added a `single_filter_stream` fast path. For standard LZMA2 blocks, data is now piped directly from the decoder to `CrcWriter`, completely bypassing the intermediate `tmpbuf: Vec<u8>` allocation.
* Updated `decode_filter` to parse the LZMA2 dictionary size from the block header properties and pass it down to the decoder.

**2. LZMA2 Layer: Activating the Circular Buffer (`src/decode/lzma2.rs`)**
* Changed `Lzma2Decoder::decompress` to instantiate `LzCircularBuffer` instead of `LzAccumBuffer`. 
* Updated `Lzma2Decoder::new()` to accept a `dict_size` parameter.
* Updated `parse_lzma` and `parse_uncompressed` to use dynamic trait objects (`impl LzBuffer<W>`) so they can accept either buffer type.

**3. Buffer Trait Refactoring (`src/decode/lzbuffer.rs`)**
* Moved `append_bytes()` and `reset()` from standalone methods on `LzAccumBuffer` into the `LzBuffer` trait.
* Implemented these trait methods for `LzCircularBuffer`. 
* *Note:* `LzCircularBuffer` already had the exact streaming logic needed (it flushes to the output writer every time the dictionary wraps around), but it was entirely unused by the decoder until now!

### Testing
Added a strict behavioral test (`test_true_streaming_decompression`) using a custom `WriteCounter` writer. 
* We compress 16KB of data with a 4KB dictionary.
* The test asserts that the output writer's `write()` method is called exactly 4 times (once per dictionary flush).
* If the old accumulating buffer was still being used, `write()` would only be called once at the very end, causing the test to fail. This guarantees O(1) memory behavior.

### Performance Impact
* **Before:** Decompressing a 1GB block required ~1GB of RAM (plus the compressed input buffer).
* **After:** Decompressing a 1GB block with a 64MB dictionary requires exactly ~64MB of RAM. Output is flushed continuously to the underlying writer.


jtmoon79:
@Hahihula could this fix the problem mentioned in #110 ?

> Decompressing large files can lead to _Out Of Memory_ errors.

jtmoon79 (approved):

Hahihula:
> @Hahihula could this fix the problem mentioned in #110 ?
> 
> > Decompressing large files can lead to _Out Of Memory_ errors.

Yes, this should fix it

alirana01 (approved):

jtmoon79:
@Hahihula this is a concerning build warning

```
$ cargo build
   Compiling crc-catalog v2.4.0
   Compiling byteorder v1.5.0
   Compiling crc v3.2.1
   Compiling lzma-rs v0.3.0 (/home/user/Projects/lzma-rs)
warning: struct `LzAccumBuffer` is never constructed
  --> src/decode/lzbuffer.rs:44:12
   |
44 | pub struct LzAccumBuffer<W>
   |            ^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: associated items `from_stream`, `append_bytes`, and `reset` are never used
  --> src/decode/lzbuffer.rs:62:12
   |
58 | / impl<W> LzAccumBuffer<W>
59 | | where
60 | |     W: io::Write,
   | |_________________- associated items in this implementation
61 |   {
62 |       pub fn from_stream(stream: W, memlimit: usize) -> Self {
   |              ^^^^^^^^^^^
...
72 |       pub fn append_bytes(&mut self, buf: &[u8]) {
   |              ^^^^^^^^^^^^
...
78 |       pub fn reset(&mut self) -> io::Result<()> {
   |              ^^^^^

warning: `lzma-rs` (lib) generated 2 warnings
```

